PR #7, opened 2020-01-01
ENH swap openmp

beckermr:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.


beckermr:

beckermr:
This build is failing for some reason that appears to be unrelated to this PR. @isuruf any ideas?

beckermr:
Is this section suspicious?

```
2020-01-01T19:08:55.3703102Z Version of gcc
2020-01-01T19:08:55.3703593Z > 1. 9.2.0 (GCC_V_9)
2020-01-01T19:08:55.3703712Z   2. 8.3.0 (GCC_V_8)
2020-01-01T19:08:55.3703917Z   3. 7.4.0 (GCC_V_7)
2020-01-01T19:08:55.3704000Z   4. 6.5.0 (GCC_V_6)
2020-01-01T19:08:55.3704080Z   5. 5.5.0 (GCC_V_5)
2020-01-01T19:08:55.3704183Z   6. 4.9.4 (GCC_V_4_9)
2020-01-01T19:08:55.3704269Z   7. 4.8.5 (OBSOLETE) (GCC_V_4_8)
```

beckermr:
OHHHHHH. I think this is a problem with the crosstool-ng dep. We cannot use 1.24 I think.

beckermr:
The section from the config above now has 7.3.0 in it and it is selected properly.

beckermr:

beckermr:

beckermr:

beckermr:

beckermr:
I guess the host trick will help this one as well @isuruf. Will fix!

beckermr:
Note to self. This linking failure is rather odd. I am not sure if it is related to the PR.

recipe/meta.yaml
```diff
@@ -201,6 +201,82 @@ outputs:
...
+      home: https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html
+      license: GPL
+
+  - name: libgomp
+    version: {{ libgomp_ver }}
+    target: {{ ctng_target_platform }}
+    script: install-libgomp.sh
+    requirements:
+      host:
+        - {{ pin_subpackage('libgcc-ng-build', exact=True) }}
```
isuruf: Since this package is only used here, why do we need separate package?

recipe/install-libgcc-no-gomp.sh
```diff
@@ -0,0 +1,9 @@
+source ${RECIPE_DIR}/install-libgcc.sh
+
+# remove and relink things
+rm -f ${PREFIX}/lib/libgomp.so
+rm -f ${PREFIX}/${CHOST}/sysroot/lib/libgomp.so
+
+# make the right links
+ln -s ${PREFIX}/lib/libgomp.so.${libgomp_ver:0:1} ${PREFIX}/lib/libgomp.so
```
isuruf:
```suggestion
ln -s ${PREFIX}/lib/libgomp.so.${libgomp_ver} ${PREFIX}/lib/libgomp.so
```

recipe/meta.yaml
beckermr: Right yeah. The first way intried to split the build of lib gcc involved running the build steps for libgomp in a separate script. This worked in terms of getting the binary packages, but then the test that compiles FORTRAN code with openmp would not pass. I couldn’t figure out why. 

Thus I decided on the approach here to see if I could get things to work. Here I actually build libgcc twice.  The first of these is the regular build with libgomp bundled. The second is same build, but then I remove the parts of libgomp I don’t need at the end. Finally, I then build the libgomp package from the regular build by copying the file. This setup is a bit more convoluted but now the FORTRAN openmp code compiles properly. 

I still need to push a few changes to the symlinks since they are not quite right in libgcc right now. Hopefully that doesn’t break anything. 

I’m more than happy to hear suggestions on how to do this better as always!

I don’t know these builds very well and this is the only way I could get it to work so far.

beckermr: Right yeah. The first way intried to split the build of lib gcc involved running the build steps for libgomp in a separate script. This worked in terms of getting the binary packages, but then the test that compiles FORTRAN code with openmp would not pass. I couldn’t figure out why. 

Thus I decided on the approach here to see if I could get things to work. Here I actually build libgcc twice.  The first of these is the regular build with libgomp bundled. The second is same build, but then I remove the parts of libgomp I don’t need at the end. Finally, I then build the libgomp package from the regular build by copying the file. This setup is a bit more convoluted but now the FORTRAN openmp code compiles properly. 

I still need to push a few changes to the symlinks since they are not quite right in libgcc right now. Hopefully that doesn’t break anything. 

I’m more than happy to hear suggestions on how to do this better as always!

I don’t know these builds very well and this is the only way I could get it to work so far.

recipe/install-libgcc-no-gomp.sh
beckermr: This suggestion is not right. I thought we wanted libgomp to point at libgomp.so.1. We then point libgomp.so.1 at either libgomp.so.1.0.0 or libgomp.so. Here libgomp_ver is 1.0.0 so you have only keep the first character.

beckermr:
We’re also going to need a mechanism to avoid mixing versions of libgcc-ng that don’t have libgomp split out from the llvm version of the openmp_impl. Otherwise code will seem to completely but will be silently linked to the wrong openmp. 

There is more than one way to do this, including another mutex package or run_constrained. We will need repo data patches too as well I think.

recipe/meta.yaml
beckermr: OK I am trying a version with no extra package, but still taking the two build approach.

beckermr:
Well the paths when gfortran is invoked look wrong:

```
2020-01-05T17:05:35.9394864Z lrwxrwxrwx 1 conda conda   16 Jan  5 17:05 $PREFIX/lib/libgomp.so -> libgomp.so.1.0.0
2020-01-05T17:05:35.9395921Z lrwxrwxrwx 1 conda conda   16 Jan  5 17:05 $PREFIX/lib/libgomp.so.1 -> libgomp.so.1.0.0
2020-01-05T17:05:35.9396538Z -rwxrwxr-x 1 conda conda 808K Jan  5 16:56 $PREFIX/lib/libgomp.so.1.0.0
```

Someone, somewhere is resetting the links.

They appear correct when libgcc-ng is built

```
2020-01-05T16:57:24.9984656Z debug the links...
2020-01-05T16:57:24.9985928Z + echo 'debug the links...'
2020-01-05T16:57:24.9992857Z + ls -lah /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1.0.0
2020-01-05T16:57:25.0032196Z lrwxrwxrwx 1 conda conda  272 Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1
2020-01-05T16:57:25.0033355Z lrwxrwxrwx 1 conda conda  276 Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1 -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1.0.0
2020-01-05T16:57:25.0034087Z -rwxr-xr-x 1 conda conda 808K Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1.0.0
2020-01-05T16:57:25.0042184Z + ls -lah /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so.1 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so.1.0.0
2020-01-05T16:57:25.0060901Z lrwxrwxrwx 1 conda conda 270 Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so
2020-01-05T16:57:25.0062210Z lrwxrwxrwx 1 conda conda 272 Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so.1 -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1
2020-01-05T16:57:25.0063489Z lrwxrwxrwx 1 conda conda 276 Jan  5 16:57 /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda_cos6-linux-gnu/sysroot/lib/libgomp.so.1.0.0 -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578240153460/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libgomp.so.1.0.0
```

beckermr:
Note to self for later, I need to check the links after conda build has packaged and then installed libgcc-ng. This could be a conda build / conda bug I suppose. I should also make all of the links relative right away so that conda doesn’t mess with them.

recipe/install-libgcc-no-gomp.sh
isuruf: Nope, we want `libgomp.so` to point at `libgomp.so.1.0.0` so that the SONAME will always be `libgomp.so.1`. Executable and other shared libraries will load `libgomp.so.1` which is going to swapped with llvm one.

recipe/meta.yaml
```diff
@@ -183,10 +183,13 @@ outputs:
...
   - name: libgcc-ng
     target: {{ ctng_target_platform }}
-    script: install-libgcc.sh
+    script: install-libgcc-no-gomp.sh
     requirements:
+      host:
+        - {{ pin_subpackage('openmp_impl', exact=True) }}
       run:
         - _libgcc_mutex * main
+        - {{ pin_subpackage("openmp_impl", max_pin='x.x') }}
```
isuruf:
```suggestion
        - {{ pin_subpackage("openmp_impl", max_pin=None) }}
```

recipe/meta.yaml
```diff
@@ -196,11 +199,61 @@ outputs:
...
+      host:
+        - {{ pin_subpackage('libgomp', exact=True) }}
+      run:
+        - _libgcc_mutex * main
+        - {{ pin_subpackage("libgomp", max_pin="x") }}
+    build:
+      string: 0_gnu
+      run_exports:
+        strong:
+          - {{ pin_subpackage("openmp_impl", max_pin="x.x") }}
```
isuruf:
```suggestion
          - {{ pin_subpackage("openmp_impl", max_pin=None) }}
```

recipe/meta.yaml
```diff
@@ -196,11 +199,61 @@ outputs:
...
         - test -f ${PREFIX}/lib/libgcc_s.so
+        - test -f ${PREFIX}/lib/libgomp.so
+        - test -f ${PREFIX}/*/sysroot/lib/libgomp.so
     about:
       summary: The GCC low-level runtime library
       home: https://gcc.gnu.org/onlinedocs/gccint/Libgcc.html
       license: GPL
 
+  - name: libgomp
+    version: {{ libgomp_ver }}
```
isuruf: This isn't meaningful. Let's keep the versioning the same as gcc.

recipe/install-libgcc-no-gomp.sh
beckermr: Ohhhhhh. I am so sorry! I must have misunderstood you. Thank you so much! Everyday I learn something new! :)

beckermr:
Alright, now all of the symlinks look correct except this one when gfortran is running its tests

```
lrwxrwxrwx 1 conda conda   29 Jan  6 13:53 $PREFIX/powerpc64le-conda_cos7-linux-gnu/sysroot/lib/libgomp.so.1 -> ../../../lib/libgomp.so.1.0.0
```

recipe/meta.yaml
```diff
@@ -239,7 +243,9 @@ outputs:
         - {{ pin_subpackage('libgomp', exact=True) }}
       run:
         - _libgcc_mutex * main
-        - {{ pin_subpackage("libgomp", max_pin="x") }}
+        - {{ pin_subpackage("libgomp", max_pin=None) }}
+      run_constrained:
+        - libgcc-ng * openmp_impl*
```
isuruf: Why do we need the 2 lines above? Seems unnecessary

recipe/meta.yaml
```diff
@@ -193,7 +197,7 @@ outputs:
         - {{ pin_subpackage("libgomp", exact=True) }}
         - {{ pin_subpackage("openmp_impl", max_pin=None) }}
     build:
-      number: {{ build_num }}
+      string: openmp_impl_h{{ PKG_HASH }}_{{ build_num }}
```
isuruf: Same here. We will only have one `libgcc-ng`.

recipe/meta.yaml
beckermr: We cannot mix `openmp_impl=4.5=0_llvm` with any version of `libgcc-ng` because the older ones will not have the symlinks for `libgomp.so.1` moved to a separate package. This will cause clobber errors when the conda environments are built and could cause silent bugs where LLVM OpenMP is not actually being used at run time.

beckermr: We could instead create another mutex package, but using build strings and run_constrained means we don't need to patch repo metadata.

recipe/meta.yaml
beckermr: See my comments above.

recipe/meta.yaml
beckermr: I have added the `run_constrained` section to the other PR as well.

isuruf: Okay, then let's do as below,
```suggestion
        - libgcc-ng {{ ctng_gcc }} *_{{ build_num | default(0) }}
```

and remove the `openmp_impl` build string.

isuruf: We can remove older `libgcc-ng` builds later.

beckermr: I'm not sure this will work. It is actually **any** previous build of libgcc-ng, even those on defaults. We cannot remove the defaults builds or control their build numbers etc to make sure the condition you suggested will always do what we intend it to do.

isuruf: I'd rather not worry about defaults and rely on conda solver using the latest build.

beckermr: I really don't think this will work and is a recipe for silent bugs. Are you sure?

beckermr: What if we rename the libgcc-ng package to something else so the solver never mixes them?

isuruf: That would definitely result in clobbering

isuruf: How about changing `_libgcc_mutex * main` to `_libgcc_mutex * conda-forge` to ensure conda-forge builds are used?

beckermr: Yep, we can change the mutex, but we only get to do that once unless we version the mutex build string too. 

Where is the mutext package made? I will need to put in a PR for that too.

isuruf: It's at https://github.com/conda-forge/_libgcc_mutex-feedstock/blob/master/recipe/meta.yaml
Yes, we can change to an exact pinning of `_libgcc_mutex`

beckermr:
Alright, I checked all three compilers the links like this appear wrong.

```
lrwxrwxrwx 1 conda conda 29 Jan  6 14:56 $PREFIX/powerpc64le-conda_cos7-linux-gnu/sysroot/lib/libgomp.so.1 -> ../../../lib/libgomp.so.1.0.0
```

I don't understand how these are getting reset. Once this aspect is fixed and I add the proper mutex, I think we will be done.

isuruf:
Why is that link wrong? What's the expected output?

beckermr:
It should link to libgomp.so.1 in PREFIX/lib.

beckermr:
Otherwise it won’t get remapped when using llvm openmp

beckermr:
It is also not what the build **ever** actually set that link to be as far as I can tell. This aspect is especially weird IIUIC.

isuruf:
`$PREFIX/powerpc64le-conda_cos7-linux-gnu/sysroot/lib/libgomp.so.1`  is in `openmp_impl` right? If so, that's fine. We can have `$PREFIX/powerpc64le-conda_cos7-linux-gnu/sysroot/lib/libgomp.so.1 -> $PREFIX/lib/libomp.so` in `llvm-openmp` variant.

Btw, conda-build will change the symlinks from absolute to relative and will use `realpath`.

beckermr:
So I wanted to avoid having llvm depend on the libgcc internals. 

I think the absolute to relative thing is the issue. If I make all of the links relative it should preserve them as is.

beckermr:
I also need to move that link to libgcc

isuruf:
@beckermr, is this ready now?

beckermr:
I want to merge the pinning and rerender first, but yeah otherwise I am happy.

recipe/install-libgcc-no-gomp.sh
```diff
@@ -0,0 +1,36 @@
...
+pushd ${PREFIX}/${CHOST}/sysroot/lib
+ln -s ../../../lib/libgomp.so.${libgomp_ver} libgomp.so.${libgomp_ver}
+ln -s ../../../lib/libgomp.so.${libgomp_ver:0:1} libgomp.so.${libgomp_ver:0:1}
+ln -s ../../../lib/libgomp.so libgomp.so
+popd
+
+# debugging
+echo "debug the links..."
+ls -lah ${PREFIX}/lib/libgomp.so*
+ls -lah ${PREFIX}/${CHOST}/sysroot/lib/libgomp.so*
```
beckermr: Additional debuging code, I am not sure if you want to remove it @isuruf.

isuruf:
Thanks @beckermr for your work here.

PR #11, opened 2020-01-07
ENH put all run exports in activation scripts

beckermr:
Checklist


beckermr:
@conda-forge-admin update team

recipe/meta.yaml
```diff
@@ -243,13 +246,10 @@ outputs:
       host:
         - {{ pin_subpackage('libgomp', exact=True) }}
       run:
-        - _libgcc_mutex {{ _libgcc_mutex }}
```
isuruf: Why remove this?

recipe/meta.yaml
```diff
@@ -128,6 +130,7 @@ outputs:
       strong:
         - {{ pin_subpackage("libgfortran-ng") }}
         - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
+        - {{ pin_subpackage("openmp_impl", max_pin=None) }}
```
isuruf: I don't know if this preferable or not. cc @mingwandroid.
Also, what do you think about renaming `openmp_impl` -> `_openmp_mutex` like the other mutexes?

isuruf: To expand on the first point, do we add this to all packages even those that don't need openmp or do we want to require package authors to explicitly declare an openmp dependency?

beckermr: I am fine renaming the mutex. I used the name you suggested initially in your comments on the openmp megathread.

We have to have a run export on openmp_impl at least the compiler activation scripts. It contains libgomp.so.1 and so conda-build is erroring because we are forcing it to be more careful about under/overlinking.

beckermr: run exports are not transitive and so it is not making it in even though it is a proper dep of libgcc-ng

recipe/meta.yaml
beckermr: It should be in the global pinning files so we do not need it. When I rerendered the recipe however, conda smithy doesn't appear to have picked up that version and put it in the ci_support yaml files.

isuruf: global pinning is used for `host`, not `run`.

beckermr: Well when I rerendered, conda smithy took it out and then the jinja would not work. I don't get why the global pinning does apply here too.

beckermr: The recipe has to be stable under rerendering, right?

recipe/meta.yaml
beckermr: To address your actual question, right now for linux we ship openmp without requiring anything extra. We can switch to requiring it I suppose, but this will break a lot of downstream builds.

recipe/meta.yaml
beckermr: this is fixed - was the pinning file

recipe/meta.yaml
```diff
@@ -187,11 +179,10 @@ outputs:
     requirements:
       host:
         - {{ pin_subpackage("libgomp", exact=True) }}
-        - {{ pin_subpackage('openmp_impl', exact=True) }}
+        - {{ pin_subpackage('_openmp_mutex', exact=True) }}
```
beckermr: This might still cause a circular dep. I am not 100% sure TBH.

recipe/meta.yaml
```diff
@@ -203,7 +203,6 @@ outputs:
         - test -f ${PREFIX}/lib/libgcc_s.so
         - test -f ${PREFIX}/lib/libgomp.so
         - test -f ${PREFIX}/*/sysroot/lib/libgomp.so
-        - test -f ${PREFIX}/*/sysroot/lib/libgomp.so.{{ libgomp_ver[0:1] }}
```
beckermr: I think we can add this test line back.

beckermr:
It looks like `libgcc-ng` did not get a run time dep on openmp_mutex like I thought the run exports here would have done.

```
TEST START: /home/conda/feedstock_root/build_artifacts/linux-ppc64le/libgcc-ng-8.2.0-hdd5993f_4.tar.bz2
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... done

## Package Plan ##

  environment location: /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578554478214/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p


The following NEW packages will be INSTALLED:

    _libgcc_mutex: 0.1-conda_forge  conda-forge
    libgcc-ng:     8.2.0-hdd5993f_4 local      

Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
export PREFIX=/home/conda/feedstock_root/build_artifacts/ctng-compilers_1578554478214/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/ctng-compilers_1578554478214/test_tmp
+ test -f /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578554478214/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/lib/libgcc_s.so
+ test -f /home/conda/feedstock_root/build_artifacts/ctng-compilers_1578554478214/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/lib/libgomp.so
Tests failed for libgcc-ng-8.2.0-hdd5993f_4.tar.bz2 - moving package to /home/conda/feedstock_root/build_artifacts/broken
WARNING:conda_build.build:Tests failed for libgcc-ng-8.2.0-hdd5993f_4.tar.bz2 - moving package to /home/conda/feedstock_root/build_artifacts/broken
WARNING conda_build.build:tests_failed(2231): Tests failed for libgcc-ng-8.2.0-hdd5993f_4.tar.bz2 - moving package to /home/conda/feedstock_root/build_artifacts/broken
TESTS FAILED: libgcc-ng-8.2.0-hdd5993f_4.tar.bz2
```

PR #13, opened 2020-01-10
Rename to _openmp_mutex

isuruf:
Checklist

Closes #11

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -46,11 +46,14 @@ outputs:
       runpath_whitelist:
         - "x86_64-conda_cos6-linux-gnu/sysroot/lib/libpthread-2.12.2.so"
     requirements:
+      host:
+        - {{ pin_subpackage("libgomp", exact=True) }}
       run:
         - {{ pin_subpackage("binutils_impl_" ~ ctng_target_platform, exact=True) }}
         # libstdcxx-ng is a runtime dep of gcc because LTO requires it.
         - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}
         - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
+        - {{ pin_subpackage("libgomp", max_pin=None) }}
     run_exports:
       # impose this requirement across the build/host boundary
       strong:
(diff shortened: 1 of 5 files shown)


recipe/meta.yaml
```diff
@@ -245,11 +250,14 @@ outputs:
       run:
         - _libgcc_mutex {{ _libgcc_mutex }}
         - {{ pin_subpackage("libgomp", max_pin=None) }}
+      run_constrained:
+        # conflict with previous name
+        - openmp_impl 9999
```
beckermr: We could just mark all of the builds from my first PR as broken.

recipe/meta.yaml
```diff
@@ -245,11 +250,14 @@ outputs:
...
     build:
       string: 0_gnu
       run_exports:
         strong:
-          - {{ pin_subpackage("openmp_impl", max_pin=None) }}
+          - {{ pin_subpackage("_openmp_mutex", max_pin=None) }}
```
beckermr: I am not sure we want/need this run export. People are supposed to depend on libgomp in the host section to insert the openmp_mutex run export. I don't think we need to make this mutex itself self-propagating.

recipe/meta.yaml
```diff
@@ -187,11 +190,12 @@ outputs:
     requirements:
       host:
         - {{ pin_subpackage("libgomp", exact=True) }}
-        - {{ pin_subpackage('openmp_impl', exact=True) }}
+        - {{ pin_subpackage('_openmp_mutex', exact=True) }}
       run:
         - _libgcc_mutex {{ _libgcc_mutex }}
+        - {{ pin_subpackage("_openmp_mutex", max_pin=None) }}
```
beckermr: libgomp in the host section should insert this in the run section for you, but being explicit is good too.

recipe/meta.yaml
isuruf: Yes, but this is for setups with `openmp_impl` already installed.

recipe/meta.yaml
isuruf: When conda-build gives a warning about `_openmp_mutex` being not in run, people might just use this instead of libgomp. Anyways, there's no harm having this here.

recipe/meta.yaml
isuruf: Yeah, just wanted to be explicit as you said this was not added in the earlier PR.

recipe/meta.yaml
beckermr: We had the run export for the llvm mutex on itself ignored, so we should make sure to put it back.

recipe/meta.yaml
beckermr: Ahhh I forgot about that! Good catch!

isuruf:
@beckermr, if you have time, please send a PR to port this over to 9.2.0

beckermr:
Ready to merge @isuruf?

PR #21, opened 2020-05-08
WIP try to split the sysroot

beckermr:
This PR attempts to split the sysroot out to a new package. It builds gcc_impl twice and keeps the sysroot the second time. I could not figure out a better way to do things so far. 

Checklist


recipe/meta.yaml
```diff
@@ -27,6 +27,24 @@ requirements:
...
+    build:
+      number: {{ build_num }}
+      detect_binary_files_with_prefix: False
+      missing_dso_whitelist:
+        - '*'
+      runpath_whitelist:
+        - "x86_64-conda_cos6-linux-gnu/sysroot/lib/libpthread-2.12.2.so"
+    requirements:
+      host:
+        - {{ pin_subpackage("libgomp", exact=True) }}
```
beckermr, line 41: We may not need this.

beckermr:
@isuruf this is ready for a first look. It’s not elegant but it works.

recipe/install-sysroot.sh
```diff
@@ -0,0 +1,234 @@
...
+fi
+
+make -C ${SRC_DIR}/.build/src/linux-* CROSS_COMPILE=${CHOST}- O=${SRC_DIR}/.build/${CHOST}/build/build-kernel-headers ARCH=${kernel_arch} INSTALL_HDR_PATH=${PREFIX}/${CHOST}/sysroot/usr ${VERBOSE_AT} headers_install
+
+if [[ ${ctng_libc} == gnu ]]; then
+  # Install libc libraries
+  pushd ${SRC_DIR}/.build/${CHOST}/build/build-libc-final/multilib
+    make -l BUILD_CFLAGS="-O2 -g -I${SRC_DIR}/.build/${CHOST}/buildtools/include" \
+            BUILD_LDFLAGS="-L${SRC_DIR}/.build/${CHOST}/buildtools/lib"           \
+            install_root=${PREFIX}/${CHOST}/sysroot install
```
isuruf, line 140: How about just not doing this install in `install-gcc.sh` and then adding the CDT based `sysroot_linux-64` as a run requirement?

beckermr: happy to try!

I don’t follow exactly which install commands in these scripts do what.

beckermr: It looks like there is another install into the sysroot below and maybe a bit above. I think we’ll have to build it a few times to check then outputs to make sure it is split properly.

beckermr:
@isuruf any idea why this failed?

isuruf:
The symlink is wrong.
```
Making absolute symlink x86_64-conda_cos6-linux-gnu/sysroot/x86_64-conda-linux-gnu -> /home/conda/feedstock_root/build_artifacts/ctng-compilers_1589059053938/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/x86_64-conda-linux-gnu relative
```

beckermr:
🤦

recipe/meta.yaml
```diff
@@ -27,6 +27,25 @@ requirements:
     - crosstool-ng >=1.23.0,<1.24
 
 outputs:
+  - name: ctng_sysroot_{{ ctng_target_platform }}
```
isuruf, line 30: What is this supposed to do?

beckermr: I am trying to get the sysroot link in one spot to be used later.

beckermr: Not sure if we need this in the end. Trying to make it work.

isuruf: Btw, we need to change `ctng_vendor` to `conda` instead of `conda_cos6` which should make the need for a symlink unnecessary.

beckermr: For sure, but I didn't want to break the current packages which have conda_cos6 lying around. It would be simpler to change that globally and then make a symlink later for compat. We can add the link to the sysroot cdt package.

isuruf: We can have the sysroot under `conda` and have symlinks for the binaries with `conda_cos6`.

beckermr: Yep exactly. This is what I meant above. I will push a change later to try this. Right now I want to let the current build go through.

PR #1, opened 2019-10-13
Add aarch64 and ppc64le

isuruf:
Checklist
* [ ] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #23, opened 2020-05-12
ENH split the sysroot

beckermr:
Checklist


beckermr:
@conda-forge-admin rerender

beckermr:
Well we have a real error now. It appears that the compiler cannot find certain declarations

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=160128&view=logs&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c&l=29754

beckermr:
@isuruf any ideas here? The header we want is in the sysroot package.

isuruf:
cc @mingwandroid, @nehaljwani, @katietz, @jjhelmus 

Issue is that https://github.com/ContinuumIO/crosstool-ng/blob/master/packages/glibc/2.12.2/0013-Implement-aligned_alloc-using-posix_memaligned.patch is not in the CDT package
I don't know what we can do here. Patch the CDT package header? Remove it?

beckermr:
Some of these patches are to the C code directly so they cannot be used at all with a CDT IIUIC.

isuruf:
It's a header file, so we can patch it if we want to.

beckermr:
Yes but I meant some of the others, like this one

https://github.com/ContinuumIO/crosstool-ng/blob/master/packages/glibc/2.12.2/0012-zic.c-use-memcpy.patch

isuruf:
Those are probably fixed in centos as well.

beckermr:
Hmmmm. So how does one know which version of glibc one has? The centos CDT is marked as 2.12. Do they have a 2.12.2 one the matches the crosstool-ng packages?

mingwandroid:
It's difficult to say what's best. In CentOS5 days, at the start of these ctng compilers I imported all the RH patches. It was 87MB if I remember correctly. I don't think there are so many these days.

@nehaljwani had some ideas about using CDT packages for the glibc and kernel headers. I support that FWIW.

isuruf:
@mingwandroid, this PR is using the CDT packages, but the `aligned_alloc` patch is not in the CDT and that's the issue.

beckermr:
BTW, we only detected this because of a test in the recipe. Unless we have tests for the rest of the patches (or write them), we won't know which ones are there unless we examine the source.

beckermr:
@isuruf The build here appears to work in general. Maybe we punt on update 2.12.2 to use a CDT and try with 2.17 instead?

isuruf:
The idea was to not build the compiler for different sysroots and change the sysroot as needed.

beckermr:
Yup. That's fine. I am saying let's go ahead and push a 2.17 CDT and then try this PR against that. If that works, then we can push a separate build to extract the current 2.12.2 CDT from ct-ng by purely splitting the recipe.

isuruf:
Sure

mingwandroid:
Good plan!

mingwandroid:
Right. I see. It's needed for C99 support unfortunately.. Roll on centos7 cdts where this should not be needed.

beckermr:
I will get to this later today or tomorrow. Spent time gardening with the kids today!

beckermr:
Well travis barfed on disk space issues, but this works!

beckermr:
Well it worked on aarch64:

On linux we get

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=163442&view=logs&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c&l=30086

```
./c_aligned: /lib64/libc.so.6: version `GLIBC_2.16' not found (required by ./c_aligned)
```

beckermr:
Do we need the centos7 docker image?

beckermr:
Ok @isuruf this works. I’m going to send a pr into the Linux-sysroot feedstock that extracts the 2.12.2 sysroot. Then we can remove the runtime constraint here and merge this.

beckermr:
@mingwandroid @isuruf any ideas on the aarch64 build here. It worked last time and now it is erroring.

beckermr:
Looks like the version of ctng was bumped

```
crosstool-ng:  1.24.0.122_gbc6a7b9e-h46ee950_0
```

to 

```
crosstool-ng:  1.24.0.126_gccbb0fb2-h46ee950_1
```

beckermr:
specifically this PR: https://github.com/conda-forge/crosstool-ng-feedstock/pull/11/files

I wonder if we are missing patches for some of the configure scripts that we need.

beckermr:
Maybe one of these:

https://github.com/conda-forge/crosstool-ng/blob/master/packages/glibc/2.17/0008-Fix-combreloc-test-BSD-grep.patch

https://github.com/conda-forge/crosstool-ng/blob/master/packages/glibc/2.17/0007-support-make4.patch

isuruf:
Can you try deleting `$BUILD_PREFIX/share/crosstool-ng/packages/glibc/2.17/*-glibc-*.patch` for aarch64?

beckermr:
Also this commit: https://github.com/beckermr/ctng-compilers-feedstock/commit/9e35fb7061baed1299073952d51c12ed1ed88c24

isuruf:
See also https://github.com/conda-forge/ctng-compilers-feedstock/pull/1#issuecomment-542727328

beckermr:
We hit an hour on drone so that's that. We'll have to use the arm server to test further.

isuruf:
Thanks. We need to add symlinks `bin/<machine>-conda_cos6-linux-gnu -> bin/<machine>-conda-linux-gnu`.

beckermr:
OK @isuruf I've put in links to the old tools for both cos6 and cos7. Do we want to add a symlink to the sysroot itself just in case?

beckermr:
BTW, drone passed on the previous job, so that build is working too.

recipe/install-gdb.sh
```diff
@@ -25,8 +25,22 @@ pushd ${PREFIX}
...
+  old_vendor="-conda_cos7-linux-gnu-"
+fi
+
+for exe in `ls ${PREFIX}/bin/*-conda-linux-gnu-*`; do
+  nm=`basename ${exe}`
+  new_nm=${nm/"-conda-linux-gnu-"/${old_vendor}}
+  if [ ! -f ${PREFIX}/bin/${new_nm} ]; then
+    ln -s ${exe} ${PREFIX}/bin/${new_nm}
+  fi
+done
```
isuruf, line 46: Can you move the above few lines to a separate script and source it here?

beckermr: Right. That seems a lot better. :)

isuruf:
Can you also bump the build number and rerender?

beckermr:
man these build number issues....

isuruf:
@conda-forge-admin, rerender

isuruf:
There's a segfault on linux-64

beckermr:
Let's try the repackaged ct-ng sysroot. We will dump cos6 eventually and so keeping this around until then seems fine.

https://github.com/conda-forge/linux-sysroot-feedstock/pull/9

beckermr:
@isuruf this works with the 2.12.2 sysroot as expected. Are you ok going with that?

isuruf:
See my comment at https://github.com/conda-forge/linux-sysroot-feedstock/pull/9/files#r429318759

isuruf:
We should figure out why there's a segfault. Otherwise, we are bound to get a segfault if we switch to 2.17

beckermr:
e.g. this one: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=164679&view=logs&jobId=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c

beckermr:
We might still be worried, but at least the tests we have won't cover further segfaults.

isuruf:
segfault was because https://github.com/conda-forge/linux-sysroot-feedstock/issues/2

beckermr:
Fortran linking error. Any ideas?

isuruf:
I think the cloberring is the issue. Not sure where it is coming from.

beckermr:
I think all of the symlinks we make around the various libraries in the scripts all assume we use lib instead of lib64 in the sysroot. So if we adjust those it should hopefully work.

beckermr:
I think we may also have adjust the openmp stuff.

beckermr:
OK so here is an offending block of code.

```
mkdir -p ${PREFIX}/${CHOST}/sysroot/lib || true
symtargets=$(find ${PREFIX}/lib -name "libgfortran*.so*")
for symtarget in ${symtargets}; do
  symtargetname=$(basename ${symtarget})
  ln -s ${PREFIX}/lib/${symtargetname} ${PREFIX}/${CHOST}/sysroot/lib/${symtargetname}
done
```

The issue here is that if we change `lib` to `lib64`, then the package will work with the new sysroot, but it won't with the old gcc builds. 

I am going to add a `run_constrained` section with the gcc_impl version unless we have a better idea.

beckermr:
Unfortunately, we'll have to mark all of the old libgfortrans as incompatible with the new sysroot. We should probably go ahead and do this for everything in the ctng feedstock.

beckermr:
We'll need to patch 2.17 sysroots for aarch64 and ppc64le for this to work.

beckermr:
Alright @isuruf, this is ready for another look!

isuruf:
I'll see if aarch64 builds work fine locally. Before merging this though, we need to fix https://github.com/conda-forge/clangdev-feedstock/blob/master/recipe/patches/0001-Find-conda-gcc-installation.patch and also the sysroot 2.17 for aarch64 and ppc64le.

beckermr:
You ok with going for the mutex approach on the sysroot packages for now? This will enable us to enable 2.17 for linux-64 too. They are setup so that we can deprecate them later. 

If we want to use them, then every occurrence of sysroot here needs to be replaced with the mutex package after we merge those PRs and cleanup the sysroot builds.

isuruf:
If you had `run` dependencies like,
```
sysroot_linux-64=2.17 -> _current_repodata_hack=1
```
and there was an empty `_current_repodata_hack=2` then, `_current_repodata_hack=1` will not be in `current_repodata.json` making `sysroot_linux-64=2.17` impossible to solve with `current_repodata.json`.

This way, we don't need an additional dependency on `sysroot_linux-64=2.12`.
I don't want to call it a mutex, because this is not that.

beckermr:
@conda-forge-admin restart ci

beckermr:
well the kernel.org certificate expired and now the build doesn't run at all

@mingwandroid do you know how to turn off the verification?

beckermr:
@conda-forge-admin rerender

beckermr:
@conda-forge-admin restart ci

beckermr:
@conda-forge-admin restart ci

beckermr:
OK @isuruf This PR is ready for a final look. I think conda build is explicitly installing the run deps in the testing phase instead of just the package because I keep getting the 2.17 sysroot. When I install binutils on linux-64 locally, I correctly get 2.12 sysroot. Thus I put in a constraint at test time in the recipe but I don't think this indicates a bug.

beckermr:
Well by luck all of the compiler builds ran and completed on all platforms!

I think this PR is ready for final review and merging!

beckermr:
cc @isuruf

recipe/meta.yaml
```diff
@@ -46,26 +46,34 @@ outputs:
...
         - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
         - {{ pin_subpackage("libgomp", max_pin=None) }}
+        - sysroot_{{ ctng_target_platform }}
     run_exports:
       # impose this requirement across the build/host boundary
       strong:
         - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
     test:
+      requires:
+        - sysroot_{{ ctng_target_platform }} {{ conda_glibc_ver }}
```
isuruf, line 69: I'm still not sure why this is needed.

beckermr: I think it is conda build. 

Here is an install of binutils in the miniforge3 docker containers

```
(base) clarence:~ beckermr$ docker run --rm -it condaforge/miniforge3
(base) root@9926e2b6839a:/# conda install binutils
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /opt/conda

  added / updated specs:
    - binutils


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    binutils-2.34              |       h53a641e_4          26 KB  conda-forge
    binutils_impl_linux-64-2.34|       h53a641e_4         9.2 MB  conda-forge
    ca-certificates-2020.4.5.1 |       hecc5488_0         146 KB  conda-forge
    certifi-2020.4.5.1         |   py37hc8dfbb8_0         151 KB  conda-forge
    conda-4.8.3                |   py37hc8dfbb8_1         3.0 MB  conda-forge
    ld_impl_linux-64-2.34      |       h53a641e_4         616 KB  conda-forge
    openssl-1.1.1g             |       h516909a_0         2.1 MB  conda-forge
    python_abi-3.7             |          1_cp37m           4 KB  conda-forge
    sysroot_linux-64-2.12      |       h3a0023d_5        30.7 MB  conda-forge
    ------------------------------------------------------------
                                           Total:        46.0 MB

The following NEW packages will be INSTALLED:

  binutils           conda-forge/linux-64::binutils-2.34-h53a641e_4
  binutils_impl_lin~ conda-forge/linux-64::binutils_impl_linux-64-2.34-h53a641e_4
  python_abi         conda-forge/linux-64::python_abi-3.7-1_cp37m
  sysroot_linux-64   conda-forge/noarch::sysroot_linux-64-2.12-h3a0023d_5

The following packages will be UPDATED:

  ca-certificates                     2019.11.28-hecc5488_0 --> 2020.4.5.1-hecc5488_0
  certifi                                 2019.11.28-py37_0 --> 2020.4.5.1-py37hc8dfbb8_0
  conda                                        4.8.2-py37_0 --> 4.8.3-py37hc8dfbb8_1
  ld_impl_linux-64                        2.33.1-h53a641e_8 --> 2.34-h53a641e_4
  openssl                                 1.1.1d-h516909a_0 --> 1.1.1g-h516909a_0


Proceed ([y]/n)? 
```

Note that it correctly grabs the 2.12 sysroot. 

Also, if I request the sysroot directly, I get

```
base) root@9926e2b6839a:/# conda install sysroot_linux-64
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /opt/conda

  added / updated specs:
    - sysroot_linux-64


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    _sysroot_linux-64_curr_repodata_hack-3|       ha9d2b57_4           5 KB  conda-forge
    ca-certificates-2020.4.5.1 |       hecc5488_0         146 KB  conda-forge
    certifi-2020.4.5.1         |   py37hc8dfbb8_0         151 KB  conda-forge
    conda-4.8.3                |   py37hc8dfbb8_1         3.0 MB  conda-forge
    kernel-headers_linux-64-3.10.0|       h3a0023d_4         945 KB  conda-forge
    openssl-1.1.1g             |       h516909a_0         2.1 MB  conda-forge
    python_abi-3.7             |          1_cp37m           4 KB  conda-forge
    sysroot_linux-64-2.17      |       h3a0023d_4        32.5 MB  conda-forge
    ------------------------------------------------------------
                                           Total:        38.9 MB

The following NEW packages will be INSTALLED:

  _sysroot_linux-64~ conda-forge/noarch::_sysroot_linux-64_curr_repodata_hack-3-ha9d2b57_4
  kernel-headers_li~ conda-forge/noarch::kernel-headers_linux-64-3.10.0-h3a0023d_4
  python_abi         conda-forge/linux-64::python_abi-3.7-1_cp37m
  sysroot_linux-64   conda-forge/noarch::sysroot_linux-64-2.17-h3a0023d_4

The following packages will be UPDATED:

  ca-certificates                     2019.11.28-hecc5488_0 --> 2020.4.5.1-hecc5488_0
  certifi                                 2019.11.28-py37_0 --> 2020.4.5.1-py37hc8dfbb8_0
  conda                                        4.8.2-py37_0 --> 4.8.3-py37hc8dfbb8_1
  openssl                                 1.1.1d-h516909a_0 --> 1.1.1g-h516909a_0


Proceed ([y]/n)? 
```

In conda-build in the CI, I am seeing it grab 2.17. I think conda build is directly installing the deps of the things it builds instead of installing the built thing and letting the solver install the deps. This would explain the behavior we see and is consistent with the idea that version is higher than features when packages are requested directly.

beckermr: The docker container here is cos6 and I was seeing errors when building with the 2.17 sysroot related to the GLIBC version in the resulting packages at link time.

isuruf: Thanks. I'll have a look tomorrow and merge.

beckermr:
bump here @isuruf

PR #27, opened 2020-06-03
ENH conflict with older builds and move links to the sysroot

beckermr:
Checklist


recipe/meta.yaml
```diff
@@ -53,13 +53,25 @@ outputs:
...
+      run_constrained:
+        - libgcc-ng !=7.3,!=9.2     # [x86_64]
+        - libstdcxx-ng !=7.3,!=9.2  # [x86_64]
+        - libgomp !=7.3,!=9.2       # [x86_64]
+        - libgcc-ng !=8.2           # [ppc64le]
+        - libstdcxx-ng !=8.2        # [ppc64le]
+        - libgomp !=8.2             # [ppc64le]
+        - libgcc-ng !=7.3           # [aarch64]
+        - libstdcxx-ng !=7.3        # [aarch64]
+        - libgomp !=7.3             # [aarch64]
```
isuruf, line 74:
```suggestion
        - libgcc-ng !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
        - libstdcxx-ng !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
        - libgomp !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
```
```

Comment (beckermr) [reply recipe/meta.yaml:74]: I forgot 5.4.0 but why not specify this list per platform? 

Doing it this way means we could never build 8.2 on linux-64 even if we wanted to, for example. We only need to mask the builds we have now per platform.

**Comment by beckermr** [comment]
BTW, I put the libgfortran symlink into the gfortran_impl package since that is the compiler that requires it. The libcxxstd one is in gcc_impl.

File: recipe/meta.yaml (hunk 1 above)
Comment (isuruf) [reply recipe/meta.yaml]: We have to make sure we conflict with defaults too. We can build 8.4.0 on linux-64 if we wanted to. No reason to go back to 8.2.

Comment (beckermr) [reply recipe/meta.yaml]: Ah right. Defaults. Thanks!

**Comment by isuruf** [comment]
> I put the libgfortran symlink into the gfortran_impl package since that is the compiler that requires it.

Can we keep it in `gcc_impl_*`? Some packages link to it via gcc due to some transitive dependencies. It shouldn't happen, but we don't want to be backwards incompatible.

**Comment by beckermr** [comment]
> Some packages link to it via gcc due to some transitive dependencies. Ideally, it shouldn't happen, but we don't want to be backwards incompatible.

🤯

**Comment by beckermr** [comment]
I am legit confused as to how those packages run since they would not get a libgfortran run export unless they used the gfortran compiler too IIUIC.

**Comment by isuruf** [comment]
> I am legit confused as to how those packages run since they would not get a libgfortran run export unless they used the gfortran compiler too IIUIC.

`-Wl,--as-needed` would make the linker analyze it and drop it from the final set of libraries.

**Comment by beckermr** [comment]
That’s a fun trick.

File: recipe/install-gcc.sh [hunk 2]
```diff
@@ -230,8 +230,25 @@ pushd ${PREFIX}/lib
...
 
-pushd ${PREFIX}/${CHOST}/sysroot/lib
-ln -sf ../../../lib/libgomp.so libgomp.so
-popd
+# make links to libs in the sysroot
+symtargets=$(find ${PREFIX}/lib -name "libstdc++*.so*")
+for symtarget in ${symtargets}; do
+  symtargetname=$(basename ${symtarget})
+  ln -s ${PREFIX}/lib/${symtargetname} ${PREFIX}/${CHOST}/sysroot/lib/${symtargetname}
+done
```
isuruf, line 238: Since we need this only for building, let's just have the symlink only for the `.so` and not `.so.*`
```suggestion
ln -s ${PREFIX}/lib/libstdc++.so ${PREFIX}/${CHOST}/sysroot/lib/libstdc++.so
```

recipe/meta.yaml
```diff
@@ -53,13 +53,21 @@ outputs:
...
         # libstdcxx-ng is a runtime dep of gcc because LTO requires it.
         - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}
         - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
         - {{ pin_subpackage("libgomp", max_pin=None) }}
         - sysroot_{{ ctng_target_platform }}
+      run_constrained:
+        - libgcc-ng !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
+        - libstdcxx-ng !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
+        - libgomp !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
+        - libgfortran-ng !=5.4,!=7.2,!=7.3,!=8.2,!=9.2
```
isuruf, line 70: We don't need this anymore.
```suggestion
```

beckermr:
I'll merge those in a bit. I am testing a fix for another build error now.

beckermr:
@isuruf I think we need to pass `-sfn` to `ln` when making links. 

Without the `-n`, the following happens

```
(base) root@18ce081c344e:~# touch a
(base) root@18ce081c344e:~# ln -s a b
(base) root@18ce081c344e:~# ln -s b c
(base) root@18ce081c344e:~# ls -lah
total 16K
drwx------ 1 root root 4.0K Jun  4 04:15 .
drwxr-xr-x 1 root root 4.0K Jun  4 04:14 ..
-rw-r--r-- 1 root root 3.1K May 11 05:36 .bashrc
-rw-r--r-- 1 root root  148 Aug 17  2015 .profile
-rw-r--r-- 1 root root    0 Jun  4 04:15 a
lrwxrwxrwx 1 root root    1 Jun  4 04:15 b -> a
lrwxrwxrwx 1 root root    1 Jun  4 04:15 c -> b
(base) root@18ce081c344e:~# readlink a
(base) root@18ce081c344e:~# readlink b
a
(base) root@18ce081c344e:~# readlink c
b
```

However with it

```
(base) root@18ce081c344e:~# ln -sn b c
(base) root@18ce081c344e:~# ls -lah
total 16K
drwx------ 1 root root 4.0K Jun  4 04:17 .
drwxr-xr-x 1 root root 4.0K Jun  4 04:14 ..
-rw-r--r-- 1 root root 3.1K May 11 05:36 .bashrc
-rw-r--r-- 1 root root  148 Aug 17  2015 .profile
-rw-r--r-- 1 root root    0 Jun  4 04:15 a
lrwxrwxrwx 1 root root    1 Jun  4 04:15 b -> a
lrwxrwxrwx 1 root root    1 Jun  4 04:17 c -> b
```

We are using a ton of links that tie together various bits from various packages and I am worried that if they are not to the exact places we say they should be, the builds will end up broken. 

Does this make sense to you?

recipe/install-gcc.sh
```diff
@@ -230,10 +230,17 @@ pushd ${PREFIX}/lib
 ln -sf libgomp.so.${libgomp_ver} libgomp.so
 popd
 
+# making these this way so conda build doesn't muck with them
+pushd ${PREFIX}/${CHOST}/sysroot/lib
+ln -sf ../../../lib/libgomp.so libgomp.so
+ln -s ../../../lib/libgomp.so.${libgomp_ver} libgomp.so.${libgomp_ver}
+ln -s ../../../lib/libgomp.so.${libgomp_ver:0:1} libgomp.so.${libgomp_ver:0:1}
```
isuruf, line 237:
```suggestion
```

beckermr: done but I had to adjust the tests too.

recipe/install-gcc.sh
```diff
@@ -230,8 +230,26 @@ pushd ${PREFIX}/lib
...
+    symtargetname=$(basename ${symtarget})
+    ln -s ${PREFIX}/lib/${symtargetname} ${PREFIX}/${CHOST}/sysroot/lib/${symtargetname}
+  done
+done
+
+symtargets=$(find ${PREFIX}/lib -name "libgfortran*.so*")
+for symtarget in ${symtargets}; do
+  symtargetname=$(basename ${symtarget})
+  ln -s ${PREFIX}/lib/${symtargetname} ${PREFIX}/${CHOST}/sysroot/lib/${symtargetname}
+done
```
isuruf, line 253:
```suggestion
for lib in libstdc++ libgfortran libatomic libquadmath libitm libvtv lib{a,l,ub,t}san; do
  ln -s ${PREFIX}/lib/${lib}.so ${PREFIX}/${CHOST}/sysroot/lib/${lib}.so
done
```

beckermr:
ok @isuruf this looks ready to merge

beckermr:
@isuruf ready to merge? 

This time, I can build this on the server and push the aarch builds to conda-forge once it is merged so you don't have to do it.

isuruf:
Before merging this, we need to make sure that older `gcc_impl_*` only use `libgcc-ng 5.4.*|7.2.*|7.3.*|8.2.*|9.1.*|9.2.*`

beckermr:
They do not. Their constraints are things like `>=7.3.0`

beckermr:
I am confused though. I thought since we moved the symlinks this would all be fine.

isuruf:
Yes, but we need to make sure they don't use the libgcc-ng from 7.5 or (9.3 if we rebuild the `libgcc-and-libstdcxx` branch)

isuruf:
The symlinks make it so that new `gcc_impl` work fine with old `libgcc-ng`. But we want to prevent old `gcc_impl` with new `libgcc-ng`.

beckermr:
> Yes, but we need to make sure they don't use the libgcc-ng from 7.5 or (9.3 if we rebuild the libgcc-and-libstdcxx branch)

Ouch. We could insert a run constraint on the gcc_impl version in this recipe that requires `gcc_impl >= ` etc.

isuruf:
That's wrong though. We want `libgcc-ng` to be used by any older gcc version, but not cross the boundary between the `conda_cos6` and `conda` divide.

isuruf:
Yes. `libstdcxx-ng`, `libgcc-ng`, `libgomp`, `libgfortran-ng`.

beckermr:
> That's wrong though. We want libgcc-ng to be used by any older gcc version, but not cross the boundary between the conda_cos6 and conda divide.

Hmmmm. The boundary is the overall gcc version IIUIC. So I think this means we want a constraint that has the explicit boundary versions in it (e.g., `libgcc-ng` has run constraint on `gcc_impl >= 7.5.0` for `linux-64`). 

This would have the effect of preventing the new libgcc-ng from being installed with any older gcc_impl, but would allow any newer one going forward.

I might be missing something though.

isuruf:
I want this recipe to work with `defaults` too which has `gcc_linux-64 = 8.2` as well. That way, they can copy this recipe and the repodata patches fixes.

beckermr:
So to do what I am saying, we'd have to exclude 8.2 as well or bump to 8.2 here.

isuruf:
Yes, see the constraint at https://github.com/conda-forge/ctng-compilers-feedstock/pull/27#issuecomment-638966122

beckermr:
Neither offers stability against them shipping an even newer version.

What if we add the old sysroot links back to these builds? This keeps them backwards compatible.

isuruf:
> What if we add the old sysroot links back to these builds? This keeps them backwards compatible.

That works, but not adding those makes it cleaner, we find out actual bugs in our metadata. If we add those back in, we'll have to keep those links forever and I'll like to remove them as soon as possible.

> Neither offers stability against them shipping an even newer version.

I hope that defaults would use our recipe for newer versions. They've been using our new recipes for clang on osx.

beckermr:
OK. If we think they will grab our changes, then let's go with this 

linux-64

```
lib* >=7.5.0,!=8.2
```

linux-aarch64

```
lib* >=7.5.0
```

linux-ppc64le

```
lib* >=8.4.0. # covers all of 7.2, 7.3, and 8.2
```

beckermr:
woops. They should be on gcc_impl in the lib* packages.

isuruf:
Yes, we shipped `libgcc-ng` on linux-64, but not `gcc_impl_*`

beckermr:
I edited my comment to make it clearer.

beckermr:
Yes, I am suggesting we add it to the new libgcc. I am happy to hotfix the old gcc_impl, but we'd need to hotfix defaults too.

isuruf:
I'd really like to have a clean recipe without these constraints. Besides the new packages have correct dependencies. It's only the older `libgcc-ng` that have incorrect dependencies.

beckermr:
OK. Do you know how to get defaults to patch their repo data?

isuruf:
We can go ahead without patching defaults and wait for them to pick it up. (If there's any issues with mixing, I'll fix the repodata of new `gcc_impl_*`.) Repo is at https://github.com/AnacondaRecipes/repodata-hotfixes/blob/master/main.py
cc @jjhelmus, @mingwandroid, @katietz

isuruf:
@beckermr, can you build the aarch64 packages?

beckermr:
They might get through. I've noticed a few drone builds passing lately. If they do not, I will build them.

beckermr:
The drone build worked for aarch64.

isuruf:
Yes, looked like it was scheduled in the 88 large core machine instead of the 192 small core machine.

beckermr:
> Yes, looked like it was scheduled in the 88 large core machine instead of the 192 small core machine.

Whoa. Can we get drone to allow us to turn that on selectively?

isuruf:
AFAIK, they don't allow that on public drone cloud.

PR #38, opened 2020-08-01
Cross compilers

isuruf:
Checklist


recipe/install-gcc.sh
```diff
@@ -226,18 +226,16 @@ popd
 
 #${PREFIX}/bin/${CHOST}-gcc "${RECIPE_DIR}"/c11threads.c -std=c11
 
-pushd ${PREFIX}/lib
-ln -sf libgomp.so.${libgomp_ver} libgomp.so
```
beckermr, line 230: I think we still need this link?

recipe/install-libgomp.sh
```diff
@@ -17,6 +17,7 @@ mkdir -p ${PREFIX}/${CHOST}
 cp -r ${tmp_dir}/sysroot ${PREFIX}/${CHOST}/sysroot
 mkdir -p ${PREFIX}/lib
 cp ${tmp_dir}/libgomp.so.${libgomp_ver} ${PREFIX}/lib/libgomp.so.${libgomp_ver}
+ln -s ${PREFIX}/lib/libgomp.so.${libgomp_ver} ${PREFIX}/lib/libgomp.so
```
beckermr, line 20: Ohhh you moved it here. Why?

isuruf: That link is the only link that would clobber installing two compilers in the same environment.

beckermr: Right because it is not in the sysroot. But why did we move it to there in the first place?

Looks like the last commit was this PR: https://github.com/conda-forge/ctng-compilers-feedstock/pull/13

recipe/meta.yaml
```diff
@@ -54,34 +54,37 @@ outputs:
     requirements:
       host:
         - sysroot_{{ ctng_target_platform }} {{ conda_glibc_ver }}
-        - {{ pin_subpackage("libgomp", exact=True) }}
-        - {{ pin_subpackage("libstdcxx-ng", exact=True) }}
-        - {{ pin_subpackage("libgcc-ng", exact=True) }}
-        - {{ pin_subpackage("libgfortran-ng") }}
```
beckermr, line 60: Why did you remove the exact pins in host here?

beckermr: In particular, if the build picks up older versions of these packages with different layouts, I am not sure if the final compiler would be right.

beckermr: Or older builds at the same version that change the layout. 

Also, I guess we are rely on build number priority to link all of the packages that have the same relative layouts? 

Seems fine but maybe exact pins would be safer?

isuruf:
@beckermr, can you review again

recipe/install-gcc.sh
```diff
@@ -182,14 +182,9 @@ $PREFIX/bin/${CHOST}-gcc -dumpspecs > $specdir/specs
 #   package is installed.
 sed -i -e "/\*link_libgcc:/,+1 s+%.*+& -rpath ${PREFIX}/lib+" $specdir/specs
 
-# Ensure that libgcc_s.so is found in the sysroot. I have done this to mask the fact that
-# strong run_export packages do not get installed into the host prefix (AFAICT) and we
-# should really fix that too. (ping @msarahan)
-cp -f ${PREFIX}/${CHOST}/lib/libgcc_s.so* ${PREFIX}/${CHOST}/sysroot/lib
```
beckermr, line 188: Is this conda install issue fixed or did we do some other hack to make this work?

recipe/meta.yaml
```diff
@@ -154,26 +188,28 @@ outputs:
...
       run:
         # For cpp and crt{i,n}.o
-        - gcc_impl_{{ ctng_target_platform }} >=7.2.0
-        - {{ pin_subpackage("libgfortran-ng") }}
-        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
+        - gcc_impl_{{ ctng_target_platform }} >={{ ctng_gcc }}
+        - libgfortran-ng {{ ctng_gcc }}.*       # [target_platform == ctng_target_platform]
         - sysroot_{{ ctng_target_platform }}
+        # Any version newer than the devtoolset works. This can be removed when bootstrapping.
+        - libgcc-ng >=4.9
```
beckermr, line 195: While this may be technically true from an ABI perspective, for the conda packages I am not sure we can do this. At some point, we changed the layout of libgcc-ng to support swapping openmp. While this change is mostly a runtime change, we may see conflicts with older versions of libgcc-ng that had libgomp vendored.

recipe/meta.yaml
```diff
@@ -54,34 +80,41 @@ outputs:
...
-        - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}
-        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
-        - {{ pin_subpackage("libgomp", max_pin=None) }}
+        - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}    # [target_platform == ctng_target_platform]
+        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}       # [target_platform == ctng_target_platform]
+        - {{ pin_subpackage("libgomp", max_pin=None) }}         # [target_platform == ctng_target_platform]
         - sysroot_{{ ctng_target_platform }}
+        # Any version newer than the devtoolset works. This can be removed when bootstrapping.
+        - libstdcxx-ng >=4.9
+        - libgcc-ng >=4.9
```
beckermr, line 97: While this may be technically true from an ABI perspective, for the conda packages I am not sure we can do this. At some point, we changed the layout of libgcc-ng to support swapping openmp. While this change is mostly a runtime change, we may see conflicts with older versions of libgcc-ng that had libgomp vendored.

recipe/install-gcc.sh
```diff
@@ -116,7 +116,7 @@ done
 exec gcc $fl ${1+"$@"}
 EOF
 
-  chmod 755 ${PREFIX}/bin/c{8,9}9
+  chmod 755 ${PREFIX}/bin/${CHOST}-c{8,9}9
```
beckermr, line 119: Should we put back links to c89 and c99 here to help downstream users? Right now I am guessing this might break some builds? Maybe only do this on the native compilers?

recipe/install-libgomp.sh
beckermr: I am still wondering if you have thoughts on this thread.

recipe/meta.yaml
```diff
@@ -31,6 +31,32 @@ requirements:
     - crosstool-ng >=1.24.0,<1.25
 
 outputs:
+  - name: libgcc-devel_{{ ctng_target_platform }}
+    script: install-libgcc-devel.sh
+    build:
+      number: {{ build_num }}
+      detect_binary_files_with_prefix: False
+      binary_relocation: False
```
beckermr, line 39: Why is all binary relocation off here?

recipe/meta.yaml
```diff
@@ -31,6 +31,32 @@ requirements:
...
+      summary: The GNU C development libraries and object files
+      home: https://gcc.gnu.org/
+      license: GPL-3.0-with-GCC-exception
+
+  - name: libstdcxx-devel_{{ ctng_target_platform }}
+    script: install-libstdc++-devel.sh
+    build:
+      number: {{ build_num }}
+      detect_binary_files_with_prefix: False
+      binary_relocation: False
```
beckermr, line 52: Again why is all binary relocation off here?

recipe/meta.yaml
```diff
@@ -54,34 +80,41 @@ outputs:
...
+        - {{ pin_subpackage("libgcc-devel_" ~ ctng_target_platform, exact=True) }}
         # libstdcxx-ng is a runtime dep of gcc because LTO requires it.
-        - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}
-        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
-        - {{ pin_subpackage("libgomp", max_pin=None) }}
+        - {{ pin_subpackage("libstdcxx-ng", max_pin=None) }}    # [target_platform == ctng_target_platform]
+        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}       # [target_platform == ctng_target_platform]
+        - {{ pin_subpackage("libgomp", max_pin=None) }}         # [target_platform == ctng_target_platform]
         - sysroot_{{ ctng_target_platform }}
+        # Any version newer than the devtoolset works. This can be removed when bootstrapping.
```
beckermr, line 95: A few questions here:

1. Why are the requirements different for native versus cross compilers here?

2. Why are we not adding a requirement on libgomp for cross compilers?

recipe/meta.yaml
```diff
@@ -154,26 +188,28 @@ outputs:
         - sysroot_{{ ctng_target_platform }}
       run:
         # For cpp and crt{i,n}.o
-        - gcc_impl_{{ ctng_target_platform }} >=7.2.0
-        - {{ pin_subpackage("libgfortran-ng") }}
-        - {{ pin_subpackage("libgcc-ng", max_pin=None) }}
+        - gcc_impl_{{ ctng_target_platform }} >={{ ctng_gcc }}
+        - libgfortran-ng {{ ctng_gcc }}.*       # [target_platform == ctng_target_platform]
         - sysroot_{{ ctng_target_platform }}
+        # Any version newer than the devtoolset works. This can be removed when bootstrapping.
```
beckermr, line 194: I have the same two questions here as I did for gcc above.

recipe/install-gcc.sh
isuruf: Yes, it is fixed. `conda-build` will install strong `run_exports` packages in `host`

recipe/meta.yaml
isuruf: That's not relevant at all. Let's take `gcc_impl_linux-aarch64` running on `linux-64`. `libgcc-ng` package is a runtime dep, but that is a `linux-64` package and is useful only for running the compiler, it is not used for libraries and executables built by the compiler because the compiler is producing arm64 code. The compiler will look at `libgcc-ng` from `host` which has the correct files

recipe/meta.yaml
isuruf: Same as before

recipe/install-gcc.sh
isuruf: I don't think that's necessary. We don't have `gcc` and only `${CHOST}-gcc`.

recipe/meta.yaml
isuruf: See my previous answer

recipe/meta.yaml
isuruf: Same as before

recipe/meta.yaml
beckermr: I don't think I was very clear. Let me try one more time. If we install an older libgcc-ng in the build env with the compiler when cross-compiling, I think we might get conflict errors for the build env when conda is installing the packages due to different layouts. I might not be following something here too.

recipe/meta.yaml
isuruf: Because that's what is used for others. The compilers are relocatable, so they don't need their prefixes changed. If there is one accidentally, we should ignore it. Otherwise, doing `make install` in outputs will trigger recompilation as the binary would change. (If `binary_relocation` is turned on and there's a prefix in the binary, `conda` will patch it for each output in a recipe and that will trigger `make` to a full rebuild. I'll try and dig up the issue.

recipe/meta.yaml
isuruf: Same as before.

recipe/install-gcc.sh
beckermr: OK. We did have these in bins on previous releases so this is a breaking changed. Happy to punt though. They should have always been prexifed by CHOST anyways.

isuruf:
@conda-forge-admin, rerender

beckermr:
@conda-forge-admin relint

(just for testing)

recipe/meta.yaml
isuruf: I've kept the requirements as is for native compilers, so there shouldn't be a problem.

beckermr: Ahhhhh. The cross compilers all have the new layout. SGTM and thanks for answering my questions!

recipe/install-libgomp.sh
isuruf: I can't remember why we did, but the link `ln -s ${PREFIX}/libgomp.so ${PREFIX}/${CHOST}/sysroot/lib/libgomp.so` is wrong because it is linking a `native` library into a cross compiling dir which is wrong because the two architectures might be different.

PR #51, opened 2021-04-12
fix not proper rpath for libgfortran's .so files

katietz:
The current libgfortran*.so* files are containing still a rpath to their original _build_env.  So we need to do the same as we do for libstdc++*.so* files.


isuruf:
Thanks @katietz. Can you bump the build number?

katietz:
> 
> 
> Thanks @katietz. Can you bump the build number?

Sure, done. Sorry, missed it

beckermr:
Merge without travis?

PR #53, opened 2021-07-15
No ctng and build GCC 10, 11

### LINKED ISSUE #55: libstdcxx-ng needs to be updated (since it's breaking appimages that are being built using conda)
https://anaconda.org/conda-forge/libstdcxx-ng needs to be updated to v6.0.29  
The problem is that the current version 6.0.28 is breaking all [AppImages](https://appimage.org/) we are building for FreeCAD. 
https://github.com/FreeCAD/FreeCAD-Bundle/issues/73#issuecomment-860801975

cc @looooo @sgrogan @probonopd

isuruf:
cc @beckermr, @chenghlee, @katietz, @ocefpaf

This is the first to step to remove crosstool-ng and build gcc directly using `./configure && make`.

The reason to remove crosstool-ng is it's hard to work with and building cross-native or canadian cross compilers is hard. (It's possible, but hard). It's also building the sysroot and binutils redundantly because we build them elsewhere.

I've only added the bare minimum to make this work. I intend to do some more refactoring like `ctng_gcc->gcc_version` , but
I've hold off on that to make the review easier. (I'll also add cross-native compilers in a separate PR)

I've also added `conda-forge/label/gcc_dev` so that these compilers are not uploaded to `main` right now, so that I can play with
them for a while. When this PR is ready to be merged, I'll make a `gcc-feedstock` through staged-recipes.

Closes https://github.com/conda-forge/ctng-compilers-feedstock/pull/50
Fixes #55


recipe/install-gcc.sh
```diff
@@ -215,26 +154,31 @@ pushd ${PREFIX}
...
   done
 popd
 
 #${PREFIX}/bin/${CHOST}-gcc "${RECIPE_DIR}"/c11threads.c -std=c11
 
+mkdir -p ${PREFIX}/${CHOST}/lib
+
 if [[ "$target_platform" == "$ctng_target_platform" ]]; then
   # making these this way so conda build doesn't muck with them
-  pushd ${PREFIX}/${CHOST}/sysroot/lib
```
beckermr, line 228: I don't understand this change in layout. Is this going to affect builds with CDTs?

isuruf: No, this is not going to affect CDTs. What this change does is, it avoids mucking with the sysroot with files from the compiler, so that only `sysroot_{{ target_platform }}` and CDTs add files to them. One advantage will be that we'll be able to swap out the conda sysroot with system one by doing
```bash
rm -rf ${PREFIX}/${CHOST}/sysroot
ln -sf / ${PREFIX}/${CHOST}/sysroot
```

beckermr: Cool. Thanks for the explanation!

beckermr (commented):
Great work! 

My main concern here is that we will get a different layout of the files between the different split packages and that will break the compilers if we mix stuff built by crosstool-ng and the gcc build.

It looks like you beefed up the tests a bit to make sure each package has the files it is supposed to have. Should we do more of that?

isuruf:
> My main concern here is that we will get a different layout of the files between the different split packages and that will break the compilers if we mix stuff built by crosstool-ng and the gcc build.

The differences are in `gcc_impl, gxx_impl, libgcc_devel, libstdcxx_devel` which pin each other exactly. So I don't think mixing is possible.
We'll be bumping the gcc minor versions `9.3.0->9.4.0` when publishing these and different versions cannot be mixed anyway. (Note that the channel_target is `conda-forge/label/gcc_dev`)

beckermr:
Ahhh great! Then as long as the compilers work, we should be good to go!

recipe/meta.yaml
```diff
@@ -1,34 +1,30 @@
...
 
 package:
-  name: ctng-compilers
-  version: {{ ctng_gcc }}
+  name: gcc_compilers
+  version: {{ version }}
 
 source:
-  path: .
+  - url: https://ftp.gnu.org/gnu/gcc/gcc-{{ version }}/gcc-{{ version }}.tar.gz
```
beckermr, line 11: We'll want the sha256 here before merging.

isuruf: Yes, will add that.

recipe/meta.yaml
```diff
@@ -399,10 +394,13 @@ outputs:
...
         - __glibc
     requirements:
+      build:
+        - binutils_impl_{{ target_platform }}  {{ ctng_binutils }}
+        - binutils_impl_{{ ctng_target_platform }}  {{ ctng_binutils }}
       host:
-        - sysroot_{{ ctng_target_platform }}
+        - sysroot_{{ ctng_target_platform }} {{ conda_glibc_ver }}
       run_constrained:
-        - libgfortran-ng {{ ctng_gcc }} *_{{ build_num | default(0) }}
```
beckermr, line 405: This was here to avoid mixing the old and new packages after the gfortran rename. I guess we don't need it anymore due to the version bump.

recipe/meta.yaml
```diff
@@ -399,10 +394,13 @@ outputs:
...
+        - libgfortran-ng {{ ctng_gcc }}
```
isuruf, line 403:
```suggestion
        - libgfortran-ng {{ ctng_gcc }} *_{{ build_num | default(0) }}
```

beckermr (commented):
Do we still want to keep the `-ng` in the package names? Maybe we remove once a new gcc feedstock is created?

isuruf:
I don't mind keeping it as it might be disruptive to remove. Besides I don't think `-ng` is about crosstool-ng, but just new generation.

recipe/build.sh
```diff
@@ -94,6 +94,7 @@ cd build
   --enable-plugin \
   --enable-gold \
   --disable-nls \
+  --disable-bootstrap \
```
beckermr, line 97: do we still want a bootstrap build when using the system compilers or should we always leave this off?

isuruf: I tried enabling it and it failed with a comparison failure between stage-2 and stage3.

With ct-ng, we didn't have the 3-stage compilation, but only 2 stages. With this PR, when using `system` compilers, it's only 1 stage and when using `gcc_bootstrap_` compilers, it's 2 stages just like ct-ng, but unlike ct-ng, we need to do build the compilers twice as conda packages which is okay IMO.

beckermr: Sounds good to me. We don't bootstrap gfortran on osx either, so at least we are consistent!

isuruf:
Let me know if you think this is good to merge and I'll fix some things up and merge.

beckermr:
LGTM!

PR #72, opened 2021-09-27
Make user given rpaths have higher priority over PREFIX

isuruf:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "gcc_impl" %}
 {% set version = gcc_version %}
 {% set chost = gcc_machine ~ "-" ~ gcc_vendor ~ "-linux-gnu-" %}
-{% set build_num = 8 %}
+{% set build_num = 9 %}
 
 package:
   name: gcc_compilers
@@ -181,6 +181,7 @@ outputs:
         - ${CC} -Wall tests/aligned_alloc.c -c -o c_aligned.o -v -fsanitize=address
         - ${CC} -Wall tests/aligned_alloc.c -c -o c_aligned.o -v
         - ${CC} -Wall c_aligned.o -o c_aligned -v && ./c_aligned   # [cross_target_platform == target_platform]
+        - ${CC} -Wall c_aligned.o -o c_aligned -Wl,-rpath,/foo && ${READELF} -d c_aligned | grep RPATH | grep "/foo:${PREFIX}/lib"  # [cross_target_platform == target_platform]
         - test -f ${PREFIX}/{{ gcc_machine }}-{{ gcc_vendor }}-linux-gnu/lib/libgcc_s.so
         - test -f ${PREFIX}/{{ gcc_machine }}-{{ gcc_vendor }}-linux-gnu/lib/libgomp.spec
         - test -f ${PREFIX}/{{ gcc_machine }}-{{ gcc_vendor }}-linux-gnu/lib/libsanitizer.spec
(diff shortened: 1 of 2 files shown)


beckermr:
I can't say I've given this topic any thought at all. In fact, if you had asked me before, I would have said it was the opposite! ;p

The only issue I can think of is if somehow a build is linking to a system lib (and working) when it should have linked to a lib in conda (but would fail in that case). I would think we'd have wanted that to fail anyways. 

I guess users of our compilers outside of conda might depend on this? I can't say either way. 

What was the issue you found that prompted this change?

beckermr:
Could accidentally vendored copies of key system libraries be more damaging with this change? Seems like yes?

beckermr:
Actually yeah. I think this might break end user environments with the sysroot packages installed.

beckermr:
Would envs with the sysroot packages still link to system libraries correctly with this change?

I had a previous comment I deleted.

isuruf:
The only change is when a user explicitly adds a rpath using `-Wl,-rpath,/foo` flag. To get the previous behaviour, they can do `-Wl,-rpath,$PREFIX/lib -Wl,-rpath,/foo`, but the new behaviour cannot be simulated with the old compilers.

> The only issue I can think of is if somehow a build is linking to a system lib (and working) when it should have linked to a lib in conda (but would fail in that case). I would think we'd have wanted that to fail anyways.

This is impossible in previous case. You always load the lib in conda.

> What was the issue you found that prompted this change?

In a project, they were building a package with our compilers and wanted to override one library in conda by adding `-L/path/outside/conda -Wl,-rpath,/path/outside/conda`, but it kept loading the conda library.

> Could accidentally vendored copies of key system libraries be more damaging with this change? Seems like yes?

Not really. User would have to add a rpath explicitly, which IMO happens rarely and those who do should know what they are doing.

> Would envs with the sysroot packages still link to system libraries correctly with this change?

I don't understand this question.

beckermr:
Hmmm. 

When we build with conda build, we add rpath flags. Won't those be treated differently too?

isuruf:
> When we build with conda build, we add rpath flags. Won't those be treated differently too?

With this change `$PREFIX/lib` will get higher precedence than `$BUILD_PREFIX/lib` (hardcoded in the compiler) which is the correct thing.

beckermr:
I am still confused on one point. It may be me and I don't want to hold up this PR if it is not needed. 

Some people install the compilers in their runtime environments and use them to build code. In this case, with this change, which copies of system libraries (eg libm) would be used to link when compiled code is run, the ones on the system or the ones in the sysroot packages?

isuruf:
The ones on the system. (The ones in the sysroot are in `$CONDA_PREFIX/$HOST/sysroot/usr/lib` which is not added by rpath ever)

isuruf:
Thanks for the review. I'm happy to chat more if you want.

recipe/meta.yaml
```diff
@@ -181,6 +181,7 @@ outputs:
         - ${CC} -Wall tests/aligned_alloc.c -c -o c_aligned.o -v -fsanitize=address
         - ${CC} -Wall tests/aligned_alloc.c -c -o c_aligned.o -v
         - ${CC} -Wall c_aligned.o -o c_aligned -v && ./c_aligned   # [cross_target_platform == target_platform]
+        - ${CC} -Wall c_aligned.o -o c_aligned -Wl,-rpath,/foo && {{ gcc_machine }}-{{ gcc_vendor }}-linux-gnu-readelf -d c_aligned | grep RPATH | grep "/foo:${PREFIX}/lib"  # [cross_target_platform == target_platform]
```
timsnyder, line 184: @isuruf Looking at this test, do you have the selector on it because you copy pasted and just left it there?  Seems like the test above is restricted to `cross == target` because it is executing a binary on target.  The rpath check should probably happen on all builds because we're tweaking the specs to do that on all configs, right?

isuruf: We probably shouldn't add the rpath change in the spec file for `cross != target`

PR #76, opened 2021-10-12
Make sure libstdc++ is not linked to accidentally

isuruf:
Some brain-dead build systems don't use LDFLAGS and might accidentally
link to libstdc++.a. Make sure there's a libstdc++.so living right next to it.

cc @183amir

Fixes https://github.com/conda-forge/ctng-compiler-activation-feedstock/issues/57


beckermr (approved):
Seems fine but idk which builds should fail or pass here.

isuruf:
> Is that setuptools? How can we fix this?

It's `libblitz`'s build system. It doesn't use LDFLAGS as it is supposed to.

PR #82, opened 2022-01-26
Readd 7.5.0

xhochy:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)


xhochy:
We're running into https://github.com/crosstool-ng/crosstool-ng/issues/1337

xhochy:
Any ideas on how to build the compilers for the arches? Emulation takes too long and cross-compiling runs into some errors.

beckermr:
We've had to build them manually in the past on our various servers. 


The new refactored build made this better but unfortunately when we moved to the new sysroot we used the compiler version to mark the boundary between compilers with and without the bundled sysroot.

beckermr:
It may be possible to patch on a mutex package to demark compilers using the bundled vs non-bundled sysroot. If we did that, we could then build 7.5 using the new builds which would be better. 

@isuruf might have a better idea too!

xhochy:
@conda-forge/ctng-compilers This is ready for review/merge.

I made the manual builds using emulation locally and uploaded them to https://anaconda.org/uwe.korn/repo (the account only contains the manual builds from this PR)

```
✓ Created gist linux_ppc64le_ctng_binutils2.31.1ctng_cpu_archpowerpc64lectng_target_platformlinux-ppc64leold_ctng_vendorconda_cos7.log
https://gist.github.com/f6ae988bc8f21a34a565da8fccaf5461
✓ Created gist linux_aarch64_ctng_binutils2.29.1ctng_cpu_archaarch64ctng_target_platformlinux-aarch64old_ctng_vendorconda_cos7.log
https://gist.github.com/a59c399c77584540b8e2346f381b9121
```

beckermr:
Thank you! I think we need @isuruf to check this one over but I'll look too.

beckermr (commented):
My C++ foo is non-existent. For the test program, is the compiler simply ignoring the various definitions in the test file because they are not used in the program?

If it is, we should write a test that uses them and then run the test program after it is built maybe?

xhochy:
> My C++ foo is non-existent. For the test program, is the compiler simply ignoring the various definitions in the test file because they are not used in the program?
> 
> If it is, we should write a test that uses them and then run the test program after it is built maybe?

They are used. The problem is that the compiler failed in the compilation phase with an internal error. They would probably be thrown out by the linker. This is though not part of the issue we want to test.

beckermr:
Awesome. LGTM!

xhochy:
Can I merge here?

beckermr:
Yup let's go for it. Feel free to copy the builds too.

isuruf:
Sorry for the late response. Can you add yourself as a maintainer for this branch or feedstock? I don't want to maintain this branch at all.

isuruf:
Ping on this @xhochy

PR #91, opened 2022-03-27
Polish specs for non-cross, tweak config for style

### LINKED ISSUE #89: Where does GCC_CONFIGURE_OPTIONS come from in build.sh?
https://github.com/conda-forge/ctng-compilers-feedstock/blob/1e2c3c8e8332027b1eb3f14bdfecdf820400976b/recipe/build.sh#L122

I see where it was defined higher in build.sh when originally added in #56.  Any objections to taking it back out @isuruf ?  I was getting confused trying to figure out where it was coming from.

### LINKED ISSUE #90: Let's configure GCC to work straight from the box
### Comment:

When we were working with the Continuum folks during the development of conda-build 3, my colleagues and I suggested rather strongly that `-I $PREFIX/include` and the 4 required linking arguments `-Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -Wl,-disable-new-dtags -L $PREFIX/lib` be built into the configuration of GCC so that when someone does a `conda install gcc` and then tries to build something with `gcc` it operates like one would expect, treating the conda environment like the 'local' include and lib path and setting things appropriately to use RPATH everywhere.

I don't recall why Ray and Michael stuck with setting `*FLAGS` in the activation scripts.  To their credit, we have been able to get very far with that.  However, to continue polishing the end-user experience, I feel that we should still continue to use the activation scripts for building packages but that their use should be optional for users who just want to compile their own code using conda to install the prerequisites of a C/C++ development environment.

Since you guys moved away from ctng, it is probably easier to do this now.  Also, it looks like LLVM/clang has the ability to define [configuration files](https://stackoverflow.com/questions/24491386/is-there-an-equivalent-to-spec-files-for-clang-llvm-and-where-can-i-find-a-refe/24638901) and accomplish the equivalent behavior as the spec-file chicanery required to get the RPATHS by default in GCC.  Again, I think it's time to make this work.

@conda-forge/ctng-compilers folks, I'm interested to hear your feedback.  I have recipe changes building locally that I'm going to Q/A a bit before opening a PR.

timsnyder:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [X] Bumped the build number (if the version is unchanged)
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [X] Ensured the license file is being packaged.

fixes #89 
fixes #90 

* change manpath to $PREFIX/man so that they are just found when
  the environment is activated
* set the pkgversion to indicate conda-forge with our buildnumber
  PKG_BUILD_STRING is set to placeholder at configure time or I would
  have used that. Maybe the hash isn't finalized yet?
* set bugurl to ctng-compilers-feedstock newissue link
* guard rpath related flags with %{!static: }
  because I think they will break static builds. Also add
  -rpath-link, -disable_new_dtags and a -L
* Add -idirafter to specs for cc1_options and cpp_options to
  put $PREFIX/include at the end of the builtin search path
  like the system packages do for /usr/local/include
* Create a copy of the 'builtin' specs called builtin.specs
  that don't have any conda-forge munging done to them. Enables
  users to pass -specs=builtin.specs to reset all of our
  spec changes if needed.
* Also, only do the spec patching when build_platform is
  target_platform.  Perhaps I really mean to guard this with
  %(cross_compile) somehow because build_platform == target_platform
  is cross during toolchain build but we want the specs
  to be be operational when the toolchain is deployed and
  not cross compiling


timsnyder:
I probably need to add some tests for the behavior of the specs but I wanted to build more of the combinations in the matrix so that I can try them out and determine if they are working the way I want in the cross builds and cross builds of non-cross toolchains.

timsnyder:
If the selectors for the tests are correct, I need to change the patching of the specs to happen when `cross_target_platform == target_platform`. I'll think through this some more after I've had some sleep.

timsnyder:
@isuruf, what do you think of this implementation?  The more I think about this, the more it makes sense to me that the specs are only done when `cross_target_platform == target_platform` but I always create `builtin.specs` so as a user of the toolchain, you can always provide `-specs=builtin.specs` and be confident that you're getting what shipped with gcc.

isuruf:
I'm not a fan of using `-idirafter` which means headers included in $PREFIX cannot override the standard headers using `#include_next` because `-I $PREFIX/include -idirafter $PREFIX/include` is the same as `-idirafter $PREFIX/include`.

How about using `%include_noerr` to include an extra spec file that is then packaged in a package `conda-gcc-specs` which means the user can opt-in instead of opt-out?

timsnyder:
Hey @isuruf.   Thanks for the feedback.  If you want to override a system header and use `#include_next` I thought you were supposed to use `-isystem` or blow away everything with `-nostdinc` (at least that's what I've read in https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html).  If you were trying to do this for a header using the Centos system gcc that lives in `/usr/include` or `/usr/local/include` you'd run into the same problem with using -I.

I was able to play around with the following tests to confirm that `-isystem` can be used to override the `-idirafter` and get the effect you require:
```sh
set -xe

cat >thing.c <<'END'
#include <a.out.h>
#include <stdio.h>

struct has_a_info{
    unsigned long a_info;
} thing;

// make sure I'm not cheating and we really do include_next the system header
#if !defined (N_MACHTYPE)
#error
#endif

void main() {
    thing.a_info = 0xdeadbeef;
    printf("%#x\n", N_MAGIC(thing));
}
END

$CC -specs=$PWD/specs -o t1 -v thing.c

[[ "$(./t1)" == "0xbeef" ]]

cat >$PREFIX/include/a.out.h <<'END'
#define N_MAGIC(exec) ((exec).a_info << 16 | 0xfeed)

#include_next <a.out.h>
END
$CC -specs=$PWD/specs -I $PREFIX/include -o t2 -v thing.c
# you will see 
# ignoring duplicate directory "/home/centos/.conda/envs/localgcc/include"
# as it is a non-system directory that duplicates a system directory
# 
test "$(./t2)" == "0xbeef" 

$CC -specs=$PWD/specs -isystem $PREFIX/include -o t3 -v thing.c
test "$(./t3)" == "0xdeadbeeffeed"
```

I had my modified specs in $PWD because the the artifacts in the CI build are almost impossible to identify when the config names get to the length of the ctng-compilers-feedstock configs but they used the sed commands that are checked into `install-gcc.sh` on this branch.

I also just picked something randomly in the gcc builtin includes that looked like it could be overridden.  `N_MAGIC` was used only because it's `#define`d inside a `#if !defined` block.

I like your idea about using `%include_noerr` so that people can opt into this.  I definitely understand how this type of thing might cause a bunch of churn for a use-case that is slightly off the reservation.  I'll rework the PR so that we do that.  I assume it's cool if I have `conda-gcc-specs` get created out of this feedstock and have them pin the same version of gcc that they were built with, right?

timsnyder:
Hey @isuruf, I've finshed reworking my PR to use `%include_noerr` and I have encountered a problem.  It isn't documented anywhere that I've seen  but if I'm reading https://github.com/gcc-mirror/gcc/blob/eed9d091de0324453fb34580f550b1ca296d6516/gcc/gcc.cc#L2376-L2379 correctly, you can't use `%include`, `%include_noerr`, or `%rename` in `$PREFIX/lib/gcc/$CHOST/${gcc_version}/specs` (the 'main' specfile).  You can only use them in a specfile passed in via `-specs`.

How important is the current addition to the specs of only `-rpath $PREFIX/lib`?  Could we not include a default specs file in `gcc_impl_` and let people opt into having all of the additions by installing `conda-gcc-specs`?  Or would you want the default behavior to be what it is now and installing `conda-gcc-specs` would replace the current specs with my expanded ones?  This would require some form of mutex package and that seems like a lot to add.

Alternatively, we could patch toplevel gcc to allow `%include_noerr` to be used in the main specfile.  The patch will need to be different across the different versions if for no other reason than the rename of `gcc/gcc.c` to `gcc/gcc.cc`.

Thoughts?

timsnyder:
I retract what I said about the mutex package being too much.  After rereading https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=mutex#openmp it doesn't seem too difficult to put the current specs in a `_gcc_specs_mutex` that gets installed with gcc_impl_ by default and provide a new `_gcc_specs_mutex` that swaps out the specs file.

Does it make sense for the cross-compilers to add `-rpath $PREFIX/lib` to things they build?  I guess the assumption is that they will have rpath set to the placeholder and then once they are installed by conda wherever that is, they will work.  I can see why you had to make sure that a cmdline passed -rpath was put earlier in the path a few months ago.

timsnyder:
I see how the `_openmp_mutex` is swapping out the libgomp.so.1 symlink but what makes conda think that `1_gnu` is newer than `1_llvm`?  Guess I need to go reread the version comparison docs.

beckermr:
We use track features. Packages with that set are lower priority. See llvm openmp package.

timsnyder:
I thought that it might have something to do with track_features but when I saw it commented out https://github.com/conda-forge/_openmp_mutex-feedstock/blob/d531c3f3caf9731e879ea1b866b0bf9ba0c7d0a3/recipe/meta.yaml#L19-L23. I stopped thinking that...

beckermr:
Oh weird! I don't recall that change! I don't know how it works anymore. Sorry!

timsnyder:
That's okay @beckermr .  I appreciate you noticing and replying with what you remember.  Maybe since he made the change to remove `track_features` from the openmp_mutex @isuruf will be able to shed some more light.

In the meantime, I've found https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#mutex-metapackages and reread it to understand how track_features works again.

timsnyder:
I think after reading and thinking about it, the reason Isuru removed track_features from openmp doesn't apply here.  Nothing should depend directly on this mutex package with the non-default track_feature pinned.  So, I went ahead and finished implementing it that way.

isuruf:
I'm sorry. I don't remember exactly why that was done and how llvm has lower priority than gnu

isuruf:
> Does it make sense for the cross-compilers to add -rpath $PREFIX/lib to things they build?

Nope

isuruf:
> Alternatively, we could patch toplevel gcc to allow %include_noerr to be used in the main specfile. The patch will need to be different across the different versions if for no other reason than the rename of gcc/gcc.c to gcc/gcc.cc.

Or use sed in the build script to patch it out.

isuruf:
> How important is the current addition to the specs of only -rpath $PREFIX/lib?

Very important. Some build systems ignore LDFLAGS when trying to see if the compiler works and bail out early. (They die with libgcc_s.so or libstdc++.so not found or too old errors)

timsnyder:
Implementing the mutex package the way I did it doesn't work because:
```
  File "/opt/conda/lib/python3.9/site-packages/conda_build/metadata.py", line 1781, in extract_single_output_text
    output = output_matches[output_index] if output_matches else ''
IndexError: list index out of range
```

That happens because someone is calling `extract_single_output_text` with unrendered recipe despite the carefully placed comment saying you shouldn't do that at https://github.com/conda/conda-build/blob/806c3f0c12fd4f0177b396887ba84e1c5f4a1e15/conda_build/metadata.py#L1754-L1759

At this point, I can:
1. unwind the jinja loop so that the metapackage should work
2. go back to having only one optional `conda-gcc-specs_{cross_target}` package and patching gcc toplevel to allow us to use `%include_noerr`.

Option number 1 seems cleaner to me because you're clearly swapping the specfile in and out with a mutex package.  However, I haven't had to maintain those and if you guys think we should patch gcc to allow use of `%include_noerr` in the toplevel specfile, I'll do it that way.

beckermr:
Yes. Jinja2 around package outputs is buggy for sure. Best to stick to simple templating and avoid ifs

beckermr:
Bcc patches sound bad to many folks. Seems cleaner to avoid them per our previous discussions with @isuruf

beckermr:
*gcc

timsnyder:
I was thinking the gcc patching wasn't the way to go until @isuruf suggested possibly using sed to patch gcc in this case https://github.com/conda-forge/ctng-compilers-feedstock/pull/91#issuecomment-1081414898 last night.

timsnyder:
I'll start working on unrolling [the loop ](https://github.com/conda-forge/ctng-compilers-feedstock/pull/91/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR190-R242) and see if that works.  I'm still going to have two outputs with the same 'name' but different 'build.string' in one recipe and I'm not sure how conda-build will handle that.

I'll also look into submitting a PR with a test for this case but I don't know who is running things over there now that Sarahan is gone.

beckermr:
You'll get good traction on conda build prs. Jannis and co are really putting energy into the ecosystem.

timsnyder:
After I unroll the loop on what I have now, and cleaning up the dependencies a bit I have something building that uses the mutex package idea with no patching.  The recipe is really ugly because the two variants are unrolled but if it works, I suppose it won't be too bad.  If it works in my local build, I'll push it.

timsnyder:
Well, what I have now starts the build but then doesn't seem to package the `gcc-specs-mutex_{cross_target}` package before it starts trying to test things and then `gcc_impl` fails to find `gcc-specs-mutex_linux-64` when it tries to test and fails.  It's strange.  I don't understand why conda-build didn't even try to package those outputs.  conda-smithy definitely sees them and added them to the README when i rererendered.

I'm going to go back to what I had where I was just adding a `%include_noerr` specfile in an optional package and patch gcc.

beckermr:
Try putting the mutex values as variables in the conda build config and then reference them in the recipe using a single output. Once you rerender this with conda build it may work around the issue.

timsnyder:
Ok.  I can do that.  I was loath to since it will double the number of times we build gcc for something that could really be put into the `ctng-compiler-activation-feedstock`.  I guess I haven't done that because I have been under the impression that we would still want `gcc_impl_*` to have the same specfile it has had up until now and I wasn't sure about having something from this feedstock depend on the activation scripts.  Maybe I'm getting ahead of myself though.  I'll try what you suggest and see if it works.

beckermr:
I think if we carefully apply selectors to the builds it might be ok? I have not tried it. You can and likely should comment out some of the build matrix while testing to speed up rerendering.

isuruf:
Can we just go to patching gcc? The net effect of using the mutex for the spec file is the same as patching.

timsnyder:
I'm fine with either implementation.  I'm pushing on both of them to see what I can get working.   The patch does seem like much less of a headache at this point.

recipe/meta.yaml
```diff
@@ -21,6 +21,9 @@ source:
       # This patch can be removed in 10.4.0 but helps with compatibility for nccl
       # https://github.com/NVIDIA/nccl/issues/494#issuecomment-881074063
       - patches/PR100102-2.a1b3484a8e6c53c8084723e3f1738d402374198e.patch      # [gcc_version == "10.3.0"]
+      - patches/old-allow-commands-in-main-specfile.patch                      # [gcc_version != "12.0.0"]
+        # FIXME this will need to be extended for >= 12.0.0 but not sure how to do with selectors
```
beckermr, line 25: I wonder if some thing like int(gcc_version.split(".")) < 12 will work? Selectors are python expressions. This won't catch dev versions but it should be fine otherwise.

timsnyder: I'll give that a try.  Thanks.

timsnyder:
The conda-gcc-specs tests are failing because of a stupid `awk` syntax typo.  I'll get that straightened out and also look into improving the selectors.

recipe/meta.yaml
timsnyder: I tried this:
```
-      - patches/old-allow-commands-in-main-specfile.patch                      # [gcc_version != "12.0.0"]
-        # FIXME this will need to be extended for >= 12.0.0 but not sure how to do with selectors
-      - patches/new-allow-commands-in-main-specfile.patch                      # [gcc_version == "12.0.0"]
+      - patches/old-allow-commands-in-main-specfile.patch                      # [int(gcc_version.split(".")[0]) <  int(12)]
+      - patches/new-allow-commands-in-main-specfile.patch                      # [int(gcc_version.split(".")[0]) >= int(12)]
```

And both selectors evaluate to `True` somehow.  I also noticed notes in the conda-build docs saying that Jinja gets applied before the selectors and I tried some other things using jinja and got both of them to not be applied.  If I define a gcc_mv in the conda_build_config.yaml and zip it into the tuple with gcc_version, I can get that to work correctly.  Since it only makes the variant names longer but hopefully doesn't impact anything else, I'm going to do that.

timsnyder:
The [`ppc64le` cross `s390x` build for version 11.2.0](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=483840&view=logs&j=8a5e49f4-6600-5903-2269-40960836f394&t=07bbfb99-0fb6-5672-030e-0e2594814808) failed with:
```
Error response from daemon: Head "[https://quay.io/v2/condaforge/linux-anvil-cos7-x86_64/manifests/latest"](https://quay.io/v2/condaforge/linux-anvil-cos7-x86_64/manifests/latest%22): received unexpected HTTP status: 500 Internal Server Error
```

I don't see controls on Azure to rerun that failing member of the matrix.  I can't remember if that's because I need to be a maintainer to see that or you guys don't enable that or Azure doesn't have that feature...  I'm going to let the rest keep running.  If quay is having some flakiness, I'd rather see if the other 63 make it through.

timsnyder:
@conda-forge/ctng-compilers this is ready for review.  I ended up creating the new `conda-gcc-specs_{{cross_target_platform}}` package that is optional.  The package is only created when `cross_target_platform == target_platform` because I don't think it makes sense to provide it for cross compilers.  However, I kept `{{cross_target_platform}}` in the name of the package to fit the naming scheme of the other packages in this feedstock.

`gcc_impl_{{cross}}` specfile now has `%include_noerr <conda.specs>` at the bottom and still has the `-rpath $PREFIX/lib` in the same place it was at the end of `*link_command:`.  I wrapped it in `!static` because I think not having it can create problems if you try to build a static library.  I also added `-disable_new_dtags` because we want `DT_RPATH` and not `DT_RUNPATH`.  I also added `builtin.specs` to the `gcc_impl` package to make it easier for someone to run `$CC -specs=builtin.specs` and get the same behavior as `$CC -specs=<($CC -dumpspecs)`.

When you install the new optional package, it provides the `conda.specs` that:
* Add `-idirafter $PREFIX/include` to the end of `*cpp_options:` and `*cc1_options:`.  Note: `cc1plus` also uses `cc1_options`.
* Add `-rpath $PREFIX/lib -rpath-link $PREFIX/lib -disable_new_dtags -L $PREFIX/lib` to the end of `*link_command:` 

The effect of the extra specs is that you can include and link libraries provided by other conda packages without explicitly passing command-line arguments to the gcc driver.  I test this behavior in https://github.com/conda-forge/ctng-compilers-feedstock/pull/91/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR208-R211

beckermr (commented):
I have a few questions. Some of the changes in the default flags need to be thought through or commented on. Usually they were out there for a reason even if it is not left as a comment in the code.

recipe/build.sh
```diff
@@ -115,10 +119,10 @@ cd build
   --disable-bootstrap \
   --disable-multilib \
   --enable-long-long \
-  --enable-default-pie \
   --with-sysroot=${PREFIX}/${TARGET}/sysroot \
   --with-build-sysroot=${BUILD_PREFIX}/${TARGET}/sysroot \
   --with-gxx-include-dir="${PREFIX}/${TARGET}/include/c++/${gcc_version}" \
-  $GCC_CONFIGURE_OPTIONS
```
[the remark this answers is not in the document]
timsnyder: https://github.com/conda-forge/ctng-compilers-feedstock/pull/91/files/cb9c88ecf7a3a87c8e5492790fb995e154022879#diff-d7075654874cb08007a21aaab3ecd4b3453a9087e7505d034d548b8938b599bcR104 is the line where `--enable-default-pie` is still in the options.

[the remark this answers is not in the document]
timsnyder: #89 has more details about the removal of `$GCC_CONFIGURE_OPTIONS`

recipe/install-conda-specs.sh
```diff
@@ -0,0 +1,31 @@
...
+    # cmdline args or FLAGS and likewise the assumptions we have about rpath are built in)
+    #
+    # THIS IS INTENDED as a safety net for casual users who just want the native toolchain to work.
+    # It is not to be relied on by conda-forge package recipes and best practice is still to set the
+    # appropriate FLAGS vars (either via compiler activation scripts or explicitly in the recipe)
+    #
+    # We use double quotes here because we want $PREFIX and $CHOST to be expanded at build time
+    #   and recorded in the specs file.  It will undergo a prefix replacement when our compiler
+    #   package is installed.
+    sed -i -e "/\*link_command:/,+1 s+%.*+& %{\!static:-rpath ${PREFIX}/lib -rpath-link ${PREFIX}/lib -disable-new-dtags} -L ${PREFIX}/lib+" $specdir/conda.specs
```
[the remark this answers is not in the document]
timsnyder: We currently have this in LDFLAGS that get installed by the binutils activation script https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/93ab56f85e1aac593fc3980fb33cacec25b01b4b/recipe/conda_build_config.yaml#L87-L96.  I'm surprised that we don't have it for the non-x86 linux platforms.

`-disable_new_dtags` (and the default behavior in newer binutils `-enable_new_dtags`) control whether `-rpath` sets  `DT_RPATH` entry in the ELF or `DT_RUNPATH`.  `DT_RPATH` and `DT_RUNPATH` are similar but slightly different.  `DT_RPATH` is highest priority path used by the dynamic loader and can not be overriden by `LD_LIBRARY_PATH` or the system config file for the dynamic linker.  This is explained in better detail in the [manpage for `ld.so(8)`](https://man7.org/linux/man-pages/man8/ld.so.8.html). 

There is also a significant difference in the behavior in that `DT_RPATH` applies to all subsequent dependencies that need to be looked up but `DT_RUNPATH` is only applied to direct dependencies and has to be set at each level of resolving.  (i.e. if you depend on a DSO, it has to also have DT_RUNPATH set).

When binutils defaulted to including the newer DT_RUNPATH in addition to the old DT_RPATH, it made everything I was building break because most of the libraries weren't also compiled with DT_RUNPATH and their dynamic dependencies wouldn't get resolved using DT_RPATH because it gets ignored when you have both DT_RPATH and DT_RUNPATH set.

I'm still scratching my head as to why the binutils-feedstock doesn't use `-disable-new-dtags` in the LDFLAGS it sets for the non-x86 linuxen.  Perhaps I shouldn't have added it into the default specfile.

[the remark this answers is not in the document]
timsnyder: I'm still looking into why `linux-ppc64le`, `linux-aarch64` and `linux-s390x` don't use `-disable_new_dtags` but `linux-32` and `linux-64` do.  It's been that way since the beginning of the current conda-forge/ctng-compiler-activation-feedstock.  I'm not sure where the older history lives.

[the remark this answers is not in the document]
timsnyder: Ray's old activation scripts in https://github.com/conda-archive/conda-recipes/tree/master/cross-compilers didn't have `-disable_new_dtags` in them for anything.  My personal fork of the old conda-recipes repo doesn't even have the `cross-compilers` recipes in it.

isuruf:
>  However, I kept {{cross_target_platform}} in the name of the package to fit the naming scheme of the other packages in this feedstock.

That's unnecessary given that it's only for native compilers.

> https://github.com/orgs/conda-forge/teams/ctng-compilers this is ready for review. I ended up creating the new conda-gcc-specs_{{cross_target_platform}} package that is optional. The package is only created when cross_target_platform == target_platform because I don't think it makes sense to provide it for cross compilers. However, I kept {{cross_target_platform}} in the name of the package to fit the naming scheme of the other packages in this feedstock.

> gcc_impl_{{cross}} specfile now has %include_noerr <conda.specs> at the bottom and still has the -rpath $PREFIX/lib in the same place it was at the end of *link_command:. I wrapped it in !static because I think not having it can create problems if you try to build a static library.

Sounds good. `static` is not for a static library. It is for a statically linked executable, which still doesn't need an rpath.

> I also added -disable_new_dtags because we want DT_RPATH and not DT_RUNPATH. I also added builtin.specs to the gcc_impl package to make it easier for someone to run $CC -specs=builtin.specs and get the same behavior as $CC -specs=<($CC -dumpspecs).

Can we not do that? It'll be basically impossible for a user to override this because it's the last argument

> Add -idirafter $PREFIX/include to the end of *cpp_options: and *cc1_options:. Note: cc1plus also uses cc1_options.

Can you explain why `-isystem` doesn't work in a comment?
    
> Add -rpath $PREFIX/lib -rpath-link $PREFIX/lib -disable_new_dtags -L $PREFIX/lib to the end of *link_command:

Can you move `-disable_new_dtags` to the front of the options so that a user given `--enable_new_dtags` can override this?

recipe/meta.yaml
```diff
@@ -187,6 +187,46 @@ outputs:
       home: https://gcc.gnu.org/
       license: GPL-3.0-only WITH GCC-exception-3.1
 
+  - name: conda-gcc-specs_{{ cross_target_platform }}                                              # [cross_target_platform == target_platform]
+    script: install-conda-specs.sh                                                                 # [cross_target_platform == target_platform]
+    build:                                                                                         # [cross_target_platform == target_platform]
+      number: {{ build_num }}                                                                      # [cross_target_platform == target_platform]
+      detect_binary_files_with_prefix: False                                                       # [cross_target_platform == target_platform]
```
isuruf, line 194: Add the following and remove all the other selectors to make it look nicer.
```suggestion
      detect_binary_files_with_prefix: False                                                       # [cross_target_platform == target_platform]
      skip: True  # [cross_target_platform == target_platform]
```

recipe/install-conda-specs.sh
```diff
@@ -0,0 +1,31 @@
...
+    # use -idirafter to put the conda "system" includes where /usr/local/include would typically go
+    # in a system-packaged non-cross compiler
+    sed -i -e "/\*cpp_options:/,+1 s+%.*+& -idirafter ${PREFIX}/include+" $specdir/conda.specs
```
timsnyder, line 21: > Can you explain why -isystem doesn't work in a comment?

@isuruf, I thought I explained why I used `-dirafter` in a comment already.  `-isystem` might work but it puts `$PREFIX/include` at the beginning of the search path and that isn't where GCC puts the system include paths when it builds in 'non-cross' mode.  If you know that specific things are going to break by using `-idirafter` instead of `-isystem`, I'll change it.  I just feel like `-idirafter` is the "right" place to put `$PREFIX/include` if we're treating it like our own `/usr/local/include` and that is what I'm trying to do.

timsnyder: No, on centos7 and I'm pretty sure most system build gcc, `/usr/local/include` and `/usr/include` are after the gcc directories:

```
-bash-4.2$ cat /etc/redhat-release 
CentOS Linux release 7.9.2009 (Core)
-bash-4.2$ echo | /usr/bin/gcc -E -Wp,-v -x c -
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/include-fixed"
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../x86_64-redhat-linux/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/gcc/x86_64-redhat-linux/4.8.5/include
 /usr/local/include
 /usr/include
End of search list.
# 1 "<stdin>"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 1 "<command-line>" 2
# 1 "<stdin>"
```

That's why I used `-idirafter`.  And, I've confirmed that even though it is at the end of `*cpp_options:` and `*cc1_options:`, passing `-isystem $PREFIX/include` will move it earlier.  I suppose I should add that as a test.

timsnyder: @isuruf I made the changes you asked for and also added tests to make sure that `-isystem $PREFIX/include` added on the cmdline moves `$PREFIX/include` to the start of the search path so that someone can intentionally override gcc libs if they need to.  I still think `-idirafter` is the correct default location for `$PREFIX/include`

recipe/install-conda-specs.sh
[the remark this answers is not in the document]
isuruf: No clue. We can add it to LDFLAGS

recipe/install-conda-specs.sh
timsnyder: @isuruf, I think I've got this straightened out with your feedback now.

recipe/meta.yaml
```diff
@@ -1,12 +1,10 @@
-{% set name = "gcc_impl" %}
 {% set version = gcc_version %}
 {% set chost = gcc_machine ~ "-" ~ gcc_vendor ~ "-linux-gnu-" %}
-{% set build_num = 14 %}
+{% set build_num = 15 %}
 
 package:
   name: gcc_compilers
   version: {{ version }}
-
```
isuruf, line 9: Revert this

timsnyder: Done.  Thanks.  That wasn't intended.

timsnyder:
@isuruf and @beckermr, as maintainers, can you rerun https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=485018&view=logs&jobId=029f0d3c-889e-5c3b-8447-1affff42892f for a transient HTTP failure or should I just trigger a complete rerun of CI with an empty commit or close & reopen?

beckermr:
I think I restarted it.

timsnyder:
Ok @isuruf, the CI is clean again after removing the spurious deletion you caught yesterday.

recipe/install-conda-specs.sh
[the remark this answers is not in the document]
timsnyder: Captured this in an issue on the activation feedstock.  https://github.com/conda-forge/ctng-compiler-activation-feedstock/issues/72

beckermr (approved):
Thanks a bunch for all of the hard work and especially the comments/explanations here. I learned a bunch and that is always good! LGTM!

We should wait for an approval from Isuru but I think we're good to go!

beckermr:
We'll need to add this to the docs. We might also add it is a dep for the compilers meta package.

isuruf:
Do you have a local build uploaded to anaconda.org that I could try?

timsnyder:
@isuruf Let me make sure that what I have locally is with the the latest version and I'll upload it for you.

timsnyder:
@isuruf I ran a local build with what we have ready to merged and uploaded the packages to `-c tsnyder`. https://anaconda.org/tsnyder/repo

timsnyder:
> We'll need to add this to the docs.

I'll look for a spot and open a PR.

timsnyder:
So, with regard to the docs, I see three options:
1. add a sentence or two to https://conda-forge.org/docs/user/faq.html#faq-compiler-metapkg, 
2. create a "Compilers & Runtimes" section in "User Documentation" similar to what we have in https://conda-forge.org/docs/maintainer/infrastructure.html#compilers-and-runtimes but focused on end-users.  I would say that everything in the maintainer section applies but focus on what end-users want to know (essentially an expanded version of the FAQ)
3. Just expand and update https://conda-forge.org/docs/maintainer/infrastructure.html#compilers-and-runtimes

@beckermr & @isuruf, let me know if you guys have thoughts on which would be best.

beckermr:
We should add to the faq now and then make an issue for an expanded user section to be addressed later IMHO.

timsnyder:
Added PR for faq at https://github.com/conda-forge/conda-forge.github.io/pull/1685 and issue to add section to the docs https://github.com/conda-forge/conda-forge.github.io/issues/1686

recipe/build.sh
```diff
@@ -115,10 +119,10 @@ cd build
   --disable-bootstrap \
   --disable-multilib \
   --enable-long-long \
-  --enable-default-pie \
```
isuruf, line 118: Why remove this?

recipe/build.sh
```diff
@@ -115,10 +119,10 @@ cd build
...
   --with-sysroot=${PREFIX}/${TARGET}/sysroot \
   --with-build-sysroot=${BUILD_PREFIX}/${TARGET}/sysroot \
   --with-gxx-include-dir="${PREFIX}/${TARGET}/include/c++/${gcc_version}" \
-  $GCC_CONFIGURE_OPTIONS
+  --with-pkgversion="conda-forge ${gcc_version}-${PKG_BUILDNUM}" \
+  --with-bugurl="https://github.com/conda-forge/ctng-compilers-feedstock/issues/new/choose"
```
isuruf, line 126: Can you only do this if `${channel_targets} == conda-forge*`?

recipe/build.sh
timsnyder: Only removed a duplicate argument.  It's still included earlier elsewhere in the configure cmdline.

recipe/build.sh
timsnyder: addressed in https://github.com/conda-forge/ctng-compilers-feedstock/pull/91/commits/676e1c520cf18b6f058bc12485ce13cb1fdc3146

timsnyder:
@isuruf, I think I've addressed the latest comments.  Are we ready to merge?

isuruf:
Looks good to me. With`-idirafter`, it cannot be moved to the front with `-I`, but it can be moved to the middle with `-isystem`.

recipe/build.sh
```diff
@@ -25,6 +28,13 @@ if [[ "$channel_targets" == *conda-forge* && "${build_platform}" == "${target_pl
   export PATH=$SRC_DIR/cf-compilers/bin:$PATH
 fi
 
+GCC_CONFIGURE_OPTIONS=()
+
+if [[ "$channel_targets" == *conda-forge* ]]; then
+  GCC_CONFIGURE_OPTIONS+=(--with-pkgversion="conda-forge ${gcc_version}-${PKG_BUILDNUM}")
```
isuruf, line 34: Can you change this to `conda-forge gcc ${gcc_version}-${PKG_BUILDNUM}` to make sure that `gcc` is in the version string to support some scripts looking for gcc in the version string?

timsnyder: Sure.  I'll open another PR to do that later this afternoon.

timsnyder: #96

PR #108, opened 2022-10-21
Add patch that was missed from the 10.4.0 release

chrisburr:
In 10.4.0 this patch was missed from the release (it was already in the release branch so I'm not sure how) which causes parsing the headers to fail when used with clang. I've been pinning 10.3.0 where needed but this is starting to get painful and cause conflicts.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set version = gcc_version %}
 {% set chost = gcc_machine ~ "-" ~ gcc_vendor ~ "-linux-gnu-" %}
-{% set build_num = 18 %}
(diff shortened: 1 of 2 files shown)


beckermr (approved):

chrisburr:
Does anyone have an idea what is going on with s390x failing?

isuruf:
Can you revert the rerender? It's a conda-smithy bug.

SylvainCorlay:
Thanks for pushing this @chrisburr. For some reason, I am sometimes getting earlier builds of gcc 10.4 which don't include that patch.

Should they be marked as broken?

Issue #95, opened 2022-04-10
 libGL Error with Ubuntu 22.04


hmaarrfk:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

It seems that Ubuntu 22.04 doesn't play well with openGL from conda-forge.

```
mamba create --name dev --channel conda-forge --override-channels matplotlib --yes
mamba activate dev
python -c "from matplotlib import pyplot as plt; plt.plot(range(5)); plt.show()"
``` 

```
libGL error: MESA-LOADER: failed to open iris: /usr/lib/dri/iris_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: iris
libGL error: MESA-LOADER: failed to open iris: /usr/lib/dri/iris_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: iris
libGL error: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: swrast
```

While the plot shows up, other hardware accelerate stuff (like using vispy) doesn't work. I didn't know where else to post this so I figured I would post it to matplotlib since it seems to have rather high visibility into these issues.

### Installed packages

```shell
# packages in environment at /home/mark/mambaforge/envs/dev:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
alsa-lib                  1.2.3                h516909a_0    conda-forge
brotli                    1.0.9                h166bdaf_7    conda-forge
brotli-bin                1.0.9                h166bdaf_7    conda-forge
bzip2                     1.0.8                h7f98852_4    conda-forge
ca-certificates           2021.10.8            ha878542_0    conda-forge
certifi                   2021.10.8       py310hff52083_2    conda-forge
cycler                    0.11.0             pyhd8ed1ab_0    conda-forge
[...]

PR #113, opened 2023-09-10
Update to gcc 13.2.0 -- Needed for Ubuntu 23.10

hmaarrfk:
See https://gcc.gnu.org/gcc-13/

Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -16,8 +16,8 @@ jobs:
         CONFIG: linux_64_cross_target_platformlinux-64gcc_machinex86_64gcc_maj_ver12gcc_version12.3.0glibc_version2.12
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
-      linux_64_cross_target_platformlinux-64gcc_machinex86_64gcc_maj_ver13gcc_version13.1.0glibc_version2.12:
-        CONFIG: linux_64_cross_target_platformlinux-64gcc_machinex86_64gcc_maj_ver13gcc_version13.1.0glibc_version2.12
+      linux_64_cross_target_platformlinux-64gcc_machinex86_64gcc_maj_ver13gcc_version13.2.0glibc_version2.12:
+        CONFIG: linux_64_cross_target_platformlinux-64gcc_machinex86_64gcc_maj_ver13gcc_version13.2.0glibc_version2.12
         UPLOAD_PACKAGES: 'True'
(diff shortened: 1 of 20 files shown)


hmaarrfk:
I feel like I'm hitting a repeat of https://github.com/conda-forge/ctng-compilers-feedstock/issues/95  for ubuntu 23.10

hmaarrfk:
Ummmm:

> For C++, construction of the global iostream objects std::cout, std::cin, etc. is now done inside the standard library, instead of in every source file that includes the <iostream> header. This change improves the start-up performance of C++ programs, but it means that code compiled with GCC 13.1 will crash if the correct version of libstdc++.so is not used at runtime. See the [documentation](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dynamic_or_shared.html#manual.intro.using.linkage.dynamic) about using the right libstdc++.so at runtime. Future GCC releases will mitigate the problem so that the program cannot be run at all with an older libstdc++.so.

Really not sure what to do about this. @h-vetinari has anything been using GCC 13 for now?

.ci_support/linux_64_cross_target_platformlinux-s390xgcc_machines390xgcc_maj_ver13gcc_version11.4.0glibc_version2.17.yaml
```diff
@@ -15,7 +15,7 @@ docker_image:
 gcc_machine:
 - s390x
 gcc_maj_ver:
-- '11'
+- '13'
```
hmaarrfk, line 18: is this change safe?

hmaarrfk:
I can confirm that this helps get thigns working for ubuntu 23.10

hmaarrfk:
thank you!

h-vetinari:
> This change improves the start-up performance of C++ programs, but it means that code compiled with GCC 13.1 will crash if the correct version of libstdc++.so is not used at runtime.

This is pretty wild IMO, but luckily it seems that [zero](https://github.com/search?q=org%3Aconda-forge+%2F%28%3Fm%29c_compiler_version%3A%5Cs%2B%5C%23%5Cs%5C%5B%28unix%7Clinux%29%5C%5D%5Cn%5Cs*-+13%2F&type=code) feedstocks are using GCC 13 so far.

hmaarrfk:
it seems that this note was present even in GCC 13.1. I believe maybe it is warning users that they must use libstdc++ built with 13+ or else face serious consequences.

I think we do that at conda forge right?

isuruf:
Shouldn't be a problem for us

hmaarrfk:
Thanks for confirming!

Issue #122, opened 2023-12-01
Fortran compiler lacks ISO_Fortran_binding.h


aradi:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

When installing a Fortran compiler (either via the `fortran-compiler` package as user or via `{{ compiler('fortran') }}` in a recipe), the C-include file `include/ISO_Fortran_binding.h` is not installed. Missing this file prevents the compilation of modern Fortran-C interfaces. It also prevents autotools from recognizing the Fortran compiler (gfortran) being F2008-compliant. For example, this seems to be the reason, why MPICH does not build the `mpi_f08` module (conda-forge/mpich-feedstock#84) preventing the usage of the modern MPI-interface in Fortran programs.

If I build gcc myself, the include file gets installed. (I use the `--enable-languages=c,c++,fortran` option).

### Installed packages

```shell
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
binutils                  2.40                 hdd6e379_0    conda-forge
binutils_impl_linux-64    2.40                 hf600244_0    conda-forge
binutils_linux-64         2.40                 hbdbef99_2    conda-forge
c-compiler                1.6.0                hd590300_0    conda-forge
fortran-compiler          1.6.0                heb67821_0    conda-forge
gcc                       12.3.0               h8d2909c_2    conda-forge
gcc_impl_linux-64         12.3.0               he2b93b0_3    conda-forge
gcc_linux-64              12.3.0               h76fc315_2    conda-forge
gfortran                  12.3.0               h499e0f7_2    conda-forge
gfortran_impl_linux-64    12.3.0               hfcedea8_3    conda-forge
gfortran_linux-64         12.3.0               h7fe76b4_2    conda-forge
kernel-headers_linux-64   2.6.32              he073ed8_16    conda-forge
ld_impl_linux-64          2.40                 h41732ed_0    conda-forge
libgcc-devel_linux-64     12.3.0             h8bca6fd_103    conda-forge
libgcc-ng                 13.2.0               h807b86a_3    conda-forge
libgfortran5              13.2.0               ha4646dd_3    conda-forge
libgomp                   13.2.0               h807b86a_3    conda-forge
libsanitizer              12.3.0               h0f45ef3_3    conda-forge
libstdcxx-ng              13.2.0               h7e041cc_3    conda-forge
[...]

PR #123, opened 2024-01-26
include ISO_Fortran_binding.h

isuruf:
Fixes https://github.com/conda-forge/ctng-compilers-feedstock/issues/122


beckermr (approved):

aradi:
Perfect, thx for the quick fix!

aradi:
@isuruf Could you please help me to understand, which packages on which platforms are affected by this fix? (Sorry, I have still problems to understand Conda in its all depths.)

After this fix, I managed to use the `ISO_Fortran_binding.h` include file when building MPICH (and the Fortran 2008 interfaces within) on almost all platforms, but not on osx_arm64. We try to understand why (conda-forge/mpich-feedstock#89), and it seems, that the include file is not present on that platform, but I have no clue, why.

Do you have any, ideas,  suggestions, which could bring us forward? (Thx in advance!)

PR #127, opened 2024-04-20
refactor

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #129, opened 2024-04-22
fix cross-compile

isuruf:
Checklist


recipe/setup_compiler.sh
h-vetinari, line 1: Assuming we want to keep the boot-strapping capabilities, this script should probably do different things based on `use_system_gcc`?

recipe/setup_compiler.sh
```diff
@@ -0,0 +1,17 @@
...
+    conda create -p $SRC_DIR/cf-compilers -c conda-forge --yes --quiet \
+      "binutils_impl_${build_platform}" \
+      "gcc_impl_${build_platform}" \
+      "gxx_impl_${build_platform}" \
+      "gfortran_impl_${build_platform}" \
+      "binutils_impl_${target_platform}=${binutils_version}" \
+      "gcc_impl_${target_platform}" \
+      "gxx_impl_${target_platform}" \
+      "gfortran_impl_${target_platform}" \
+      "sysroot_${target_platform}=${c_stdlib_version}"
```
h-vetinari, line 13: For consistency with CBC, you should use `sysroot_version` here.
```suggestion
      "${sysroot_name}_${target_platform}=${sysroot_version}"
```

Or we rewrite CBC so that `sysroot_{name,version}` become `c_stdlib{,_version}` (but given the current impossibility to extend an existing zip in conda-build, the former is definitely easier).

recipe/setup_compiler.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [[ ! -d  $SRC_DIR/cf-compilers ]]; then
+    conda create -p $SRC_DIR/cf-compilers -c conda-forge --yes --quiet \
```
h-vetinari, line 4: I'd prefer not to use `--quiet`. It's very unusual to populate conda environments through other means than `meta.yaml`, and if it doesn't even show up in the logs, it's very hard to understand for someone looking at the recipe. Alternatively, we could add a `set -x` at the top.

isuruf: I prefer `--quiet` as opposed very verbose builds. This has been there for a long time and I prefer to keep it that way.

recipe/setup_compiler.sh
isuruf: Let's remove `use_system_gcc`

recipe/setup_compiler.sh
h-vetinari: Both resolve to
```
sysroot_linux-64=2.12
sysroot_linux-aarch64=2.17
[etc.]
```
Could you explain how they're different conceptually?

recipe/setup_compiler.sh
h-vetinari: > This has been there for a long time

Huh? You're moving build deps from `meta.yaml` into a single function call in the build scripts.

The log prominently showed the build env before anything starts, and now that information would be much much less visible.

recipe/setup_compiler.sh
isuruf: Yes, for now, they do. `$sysroot_name` is for `cross_target_platform`. `c_stdlib` is for `target_platform`.

recipe/setup_compiler.sh
isuruf: It was already there in https://github.com/conda-forge/ctng-compilers-feedstock/blob/main/recipe/build.sh#L10

isuruf: I just extended the call to all scenarios

recipe/setup_compiler.sh
isuruf: I'll rename for better clarity

recipe/setup_compiler.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [[ ! -d  $SRC_DIR/cf-compilers ]]; then
+    conda create -p $SRC_DIR/cf-compilers -c conda-forge/label/sysroot-with-crypt -c conda-forge --yes --quiet \
```
h-vetinari, line 4: So I tried to look for where this label got built, but AFAICT it just got added after the fact to a package in `main`, i.e. the last builds before https://github.com/conda-forge/linux-sysroot-feedstock/pull/54 / https://github.com/conda-forge/linux-sysroot-feedstock/pull/55. 

I saw you already [answered](https://github.com/conda-forge/ctng-compilers-feedstock/pull/127#discussion_r1573853024) a question about this in the last PR, but leaving this as a review note for posterity, because it's an unusual situation (i.e. labels normally get created in-feedstock, which isn't the case here)

recipe/meta.yaml
```diff
@@ -33,15 +33,6 @@ build:
 
 requirements:
   build:
-    - binutils_impl_{{ target_platform }}  {{ binutils_version }}
-    - binutils_impl_{{ cross_target_platform }}  {{ binutils_version }}
-    - gcc_impl_{{ target_platform }}  {{ gcc_version }}               # [build_platform != target_platform]
-    - gxx_impl_{{ target_platform }}  {{ gcc_version }}               # [build_platform != target_platform]
```
h-vetinari, line 39: Now that we're not requiring the exact same `gcc_version` for cross-compiling anymore, that should also mean that we don't have to merge with failing CI for new GCC versions, which is great 🥳

isuruf:
@conda-forge-admin, rerender

recipe/setup_compiler.sh
isuruf: Yes, I added it later. Not sure what else to do.

recipe/meta.yaml
```diff
@@ -61,17 +54,8 @@ outputs:
         - gxx_impl_{{ build_platform }}  {{ gcc_version }}                # [build_platform != target_platform]
```
h-vetinari, line 54: I think you can delete this now:
```
      ignore_run_exports_from:
        - gcc_impl_{{ build_platform }}  {{ gcc_version }}                # [build_platform != target_platform]
        - gxx_impl_{{ build_platform }}  {{ gcc_version }}                # [build_platform != target_platform]
```

recipe/setup_compiler.sh
h-vetinari: I mean if we want to, we could add branches in `linux-sysroot` corresponding to the state before the crypt-removal, add a `channel_targets` in CBC, and push those branches. That way it'd be easy to discover where the builds come from.

isuruf: Not worth it, since it's not needed in gcc 14 forwards.

h-vetinari: Fine by me. Could you leave a comment that we can remove the extra channel once we have GCC compilers?

Issue #126, opened 2024-04-19
tzdb support in libstdcxx 14


h-vetinari:
GCC 14 finally finished `<chrono>` support (search for P0355R7 [here](https://en.cppreference.com/w/cpp/compiler_support)).

We'll need to set `_GLIBCXX_ZONEINFO_DIR=$PREFIX/share/zoneinfo` (see [here](https://github.com/gcc-mirror/gcc/blob/9f295847a9c32081bdd0fe908ffba58e830a24fb/libstdc%2B%2B-v3/src/c%2B%2B20/tzdb.cc#L1062-L1083)), to point to our own [`tzdata`](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fnoarch%2Ftzdata-2024a-h0c530f3_0.conda), assuming relocation works correctly for the embedded path there. Otherwise we could patch the `zoneinfo_dir_override` function to directly load the path from `$CONDA_PREFIX`.

PR #133, opened 2024-05-07
GCC 14.1, 13.3 & 12.4

h-vetinari:
Freshly released: [14.1](https://gcc.gnu.org/pipermail/gcc/2024-May/243921.html), [13.3](https://gcc.gnu.org/pipermail/gcc/2024-May/243980.html), [12.4](https://gcc.gnu.org/pipermail/gcc/2024-June/244173.html)


h-vetinari:
@conda-forge/ctng-compilers, this is ready for review! Point releases for the maintenance branches should show up towards the end of May.

h-vetinari:
Ah, we might want to fix #126 here already.

h-vetinari:

h-vetinari:
Not sure what's up after #130, but I cannot rerender locally anymore.
```
  File "E:\miniforge\envs\builder\Lib\site-packages\conda_smithy\feedstock_io.py", line 55, in write_file
    with io.open(filename, "w", encoding="utf-8", newline="\n") as fh:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\[xxx]\\dev\\conda-forge\\ctng-compilers-feedstock\\.ci_support\\linux_aarch64_cross_target_platformlinux-64cross_target_stdlibsysrootcross_target_stdlib_version2.12gcc_maj_ver11gcc_version11.4.0old_tripletx86_64-conda_cos6-linux-gnutripletx86_64-conda-linux-gnu.yaml'
```

Might be running into path length limitations on windows...

At least the bot still can. :)

h-vetinari:
@isuruf, unsurprisingly, the mingw patches from #130 fail against GCC 14. What would you like to do here - carry two sets of patches, or build mingw only against GCC 13 (or only against 14)?

isuruf:
We can carry two sets in `recipe/patches/mingw/13` and `recipe/patches/mingw/14`

h-vetinari:
@conda-forge/ctng-compilers, this should be ready for review. We might want to link the C++ chrono implementation to a tzdb we supply ourselves (with `tzdata`; see #126), but for now this is a vanilla upgrade.

h-vetinari:

h-vetinari:
This now also contains 13.3 & the tzdb fix. PTAL @conda-forge/ctng-compilers :)

isuruf:
Thanks. 13.3 build fails to apply the patches though.

isuruf:
For tzdb, can you add a test? Binary prefix relocation usually don't work when the prefix is stored in a std::string.

h-vetinari:
Linter needs https://github.com/conda-forge/conda-smithy/pull/1946 (plus the standard `{% if gcc_maj_ver is not defined %}` dance).

h-vetinari:
> For tzdb, can you add a test? Binary prefix relocation usually don't work when the prefix is stored in a std::string.

This now has a custom `zoneinfo_dir_override` that loads from the respective environment variables dynamically, so no relocation necessary to catch the right tzdb-location. I've also adapted some upstream tests to double-check that things work.

PTAL @conda-forge/ctng-compilers

isuruf:
Thanks for adding tests and also doing the change. I'm not a fan of using env variables here. It's the only thing that depends on env variables and if we could avoid it, that should be better.

Question: where does this code end up in? `libstdc++.so` or `g++` ?

h-vetinari:
> I'm not a fan of using env variables here. It's the only thing that depends on env variables and if we could avoid it, that should be better. (One idea is to use relative paths)

I think using `zoneinfo_dir_override` is how GCC envisions these overrides. I've kept it as a weak symbol so that users of the conda-forge packages can still override it themselves.

I find `$PREFIX/share/zoneinfo` easier to follow and less brittle than relative paths, but ultimately I don't care that much how we get this done. It's early days for `<chrono>` support, so I think already that it works is a bonus, much less that we can iterate on the technical details if we want.

Overall though, I think the env var approach is the most portable/uniform, because we'll also need this for libcxx (and I also recently added this in [orc](https://github.com/apache/orc/pull/1882) to solve the same problem for windows).

> Question: where does this code end up in? `libstdc++.so` or `g++` ?

With "code" you mean the `zoneinfo_dir_override` implementation? I'm not very familiar with weak symbols, but AFAIU it's still part of the library.

h-vetinari:
> It's the only thing that depends on env variables and if we could avoid it, that should be better. (One idea is to use relative paths)

I think this is the first time that the C++ stdlib depends on external resources, so we don't really have something to compare to here. Regarding relative paths, those won't work (well) across the board AFAICT, because the headers are at different depth relative to `$PREFIX` between unix & windows (due to extra `Library/` on windows) - which isn't so much an issue for libstdcxx, but will be needed for tzdb-support on windows in other places (e.g. libcxx, orc). IMO it would be good to have a consistent approach.

Given that these environment variables are always there in conda-forge (and we [rely](https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/a9b4929bf211696131aabefe02a34f3b75f6ee11/recipe/activate-g%2B%2B.sh#L85-L93) on them in the compiler activation already), I don't see the downsides of relying on them for the tzdb as well. In any case, we still have at least a year before GCC 14 becomes the default compiler version in conda-forge, and we could use that time to refine the setup (also seeing how things go with libcxx 19 etc.).

If you're still dead set against env vars, can we merge this PR without any tzdb/tzdata bits, just the vanilla version upgrade?

h-vetinari:
@isuruf, please pick one:
1. move forward with 13.3, park 14.1 and figure out tzdb stuff later
1. use env vars for tzdb support (at least for now, can be refined later)
1. explain alternative implementation to env vars
    * if relative paths, respond to shortcomings/compatibility concerns mentioned above
1. other approach, with explanation

isuruf:
@conda-forge-admin, rerender

isuruf:
So, this code ends up in `libstdc++.so` which is in almost all envs. Using CONDA_PREFIX/PREFIX env variable will lead to hard to debug issues when an executable used while the conda environment is activated or not. Most of the important conda packages work without environment activation, and I'd like to keep it that way.
Another issue is with statically linked executables like `micromamba` which will now have different behaviour based on what CONDA_PREFIX is.

Let's remove the patch and merge and figure out tzdb stuff later.

h-vetinari:
> Most of the important conda packages work without environment activation, and I'd like to keep it that way.

That's fair, though they _would_ still keep working, because by default libstdcxx will just go look into `/usr/share/zoneinfo`, and we've kept that fallback for unactivated environments. I've added a test to this effect.

> Another issue is with statically linked executables like `micromamba` which will now have different behaviour based on what CONDA_PREFIX is.

Same here.

> Using CONDA_PREFIX/PREFIX env variable will lead to hard to debug issues when an executable used while the conda environment is activated or not.

I'm highly doubtful of this surfacing anywhere. _A_ tzdb will always be found on a unix system, so the only issues then would have to be caused by divergences between the system tzdb and what we have in `tzdata` (completely aside from the fact that C++20 `<chrono>` isn't used anywhere yet).

> Let's remove the patch and merge and figure out tzdb stuff later.

GCC 12.4 will be released on June 20th, so I propose to wait until then in any case. If you are willing to consider the above arguments (and tests), I'd prefer to move forward as-is, but happy to drop the patch if that's the cost of moving forward.

isuruf:
> A tzdb will always be found on a unix system, so the only issues then would have to be caused by divergences between the system tzdb and what we have in tzdata (completely aside from the fact that C++20 <chrono> isn't used anywhere yet).

I didn't mean that tzdb will not be found. I meant tzdb divergence in all of my arguments.

>  (completely aside from the fact that C++20 <chrono> isn't used anywhere yet).

I don't understand this. It's not used anywhere *yet*, but it will be in the future.

h-vetinari:
> I meant tzdb divergence in all of my arguments.

Can you outline a case where that would ever matter? To me this is completely hypothetical. For example, if a timezone name isn't found the code will error clearly. Yes, it's conceivable to get shifts by 1h (or a missing leap second) under extremely narrow circumstances, but I don't see how that falls under "hard to debug", because the obvious source is the tzdb (aside from the fact that C++20 provides `chrono::remote_version`, which allows authors to sanity-check the tzdb version in-code).

> It's not used anywhere yet, but it will be in the future.

My point is that usage is at zero and there will be a very gradual adoption curve, which would allow us to respond to any eventual issues. The bug surface you're pointing out is IMO so narrow (someone has an outdated system tzdb, wants to use our libstdcxx builds without any activation, has `<chrono>` code, _and_ manages to find a tzdb divergence to trip over) that I doubt it would be hit in practice.

I also wouldn't be arguing here if I saw a better solution; but we don't have constant depth of the C++ stdlib relative to `$PREFIX` across platforms, so IMO relative paths are either broken in some places, or more fragile than env vars once we need to start probing several directory levels.

isuruf:
> I don't see how that falls under "hard to debug", because the obvious source is the tzdb

That might be obvious to you right now, but it will almost always be hard to debug for the user.

> but we don't have constant depth of the C++ stdlib relative to $PREFIX across platforms

It's always in $PREFIX/lib/libstdc++.so.6

h-vetinari:
> It's always in $PREFIX/lib/libstdc++.so.6

What I mean is that on windows it's [under](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibstdcxx-13.2.0.2-h57928b3_1.conda)
```
%PREFIX%/Library/bin/libstdc++-6.dll
         ^^^^^^^^ 
         extra level compared to $PREFIX/lib/libstdc++.so.6
```
so `../../share/zoneinfo` will not work where - in some way - it's "most" important to get this right, because we don't have the fallback to `/usr/share/zoneinfo` there. The same problem applies to libc++. Ideally, I'd like to have a solution that is capable of fixing this issue for both unix & windows uniformly.

Of course, you can say that we don't need to care about non-default C++ stdlibs on windows, but then we probably shouldn't publish them (rather than leaving `<chrono>` broken)? And even then there's things like orc, which [want](https://github.com/apache/orc/blob/main/c%2B%2B/src/Timezone.cc) the tzdb despite not using C++20 yet (and where the env var approach was good enough as a fallback for upstream to merge).

isuruf:
> What I mean is that on windows it's [under](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Flibstdcxx-13.2.0.2-h57928b3_1.conda)

That's a simple `#ifdef _WIN32` .

h-vetinari:
OK, if you prefer platform-specific implementations over environment variables, that's fine by me. 🤷

h-vetinari:
> OK, if you prefer platform-specific implementations over environment variables, that's fine by me. 🤷

So I tried looking at this, and it's not trivial (from what I can see) to get the path to the library where the function lives. It'd be a bit easier to get the path of the calling executable, but that's useless because it has unknown relationship to `$PREFIX`. C++20's `std::source_location` also doesn't help. The best-looking solution I found is the small [whereami](https://github.com/gpakosz/whereami) library (particularly `wai_getModulePath`), but copying/vendoring that seems like major overkill for something that's solved in a handful of lines when using environment variables...

Or could you tell me how you you envisioned getting a path at a stable/predictable depth in the environment file hierarchy from the C++ side?

isuruf:
See https://github.com/gcc-mirror/gcc/blob/9a76db24e044c8058497051a652cca4228cbc8e9/libsanitizer/sanitizer_common/sanitizer_dl.cpp#L25-L35 for an example.

h-vetinari:
> See https://github.com/gcc-mirror/gcc/blob/9a76db24e044c8058497051a652cca4228cbc8e9/libsanitizer/sanitizer_common/sanitizer_dl.cpp#L25-L35 for an example.

Thanks. I found the dladdr approach too, but obviously that won't work for windows, where we'd have to use `GetModuleFileNameW` or similar.

h-vetinari:
So I implemented something for the relative path approach here, but obviously using `dladdr` on linux means we need to link `libstd++` to `libdl`. For now I've just hacked this with `LDFLAGS` (and even that I cannot get to work once something links to `libstd++.so` itself), but what I'd like to do is use the equivalent of CMakes `target_link_libraries` for `libstd++.so` itself.

I _think_ it should be something like `AC_LIB_LINKFLAGS`, but the [makefile](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.1.0/libstdc%2B%2B-v3/src/Makefile.in)(s) in GCC are just indecipherable for me (though I really tried to find where the shared lib gets defined...). Any ideas how/where to do this?

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++            break;
++          }
++        }
++        // <string> is included through <chrono>
++        std::string tz_dir(this_lib);
++        tz_dir += TO_PREFIX;
++        tz_dir += "/share/zoneinfo";
++        // std::string constructor deep-copies
++        free(this_lib);
++        return tz_dir.c_str();
```
mbargull, line 90: This pointer outlives the memory it points to when it is being returned.
We should return a pointer to statically allocated memory instead, I think.
Additionally, the allocations from `this_lib` and `tz_dir` are not really needed.
For `this_lib` you can just continue to use `resolved` instead.
And if you allocate `buffer` with size `PATH_MAX + sizeof(TO_PREFIX "/share/zoneinfo"` you can append the suffix directly to `resolved` in `buffer`'s memory then.

h-vetinari: Thanks! For now, I'm just trying to get something working[^1]. Improving the implementation can be done afterwards (even if it happens as part of this PR after all).

[^1]: despite trying to keep the implementation minimal/simple, I remain strongly doubtful that the trade-off of all that complexity just for avoiding a corner³ case compared to the env var approach is beneficial.

mbargull: > Improving the implementation can be done afterwards (even if it happens as part of this PR after all).

This was just me seeing a use-after-free error, so wanted to make sure that we don't include it in lower level libs.

> I remain strongly doubtful that the trade-off of all that complexity just for avoiding a corner case compared to the env var approach is beneficial.

`<chrono>` use and in conjunction with hitting tzdb absence/differences is a corner case as of now, yes.
Not having `CONDA_PREFIX` set or set to a different path is really common, though (e.g., think about `conda-build`: we have a `host` and `build` environment -- only packages from one of them can use `CONDA_PREFIX` at runtime).
There are plethora of ways things can go wrong with reliance on env vars (and with reliance on activation scripts even more so) -- it's too brittle for us to rely on in lower level libs.
(Wcan chat about why/where there are so many problems with that if things are unclear, if you want.)

h-vetinari: > Not having `CONDA_PREFIX` set or set to a different path is really common, though (e.g., think about `conda-build`: we have a `host` and `build` environment -- only packages from one of them can use `CONDA_PREFIX` at runtime).

It's easy to probe all of `CONDA_PREFIX` resp. `PREFIX` / `BUILD_PREFIX` (if `CONDA_BUILD` is set), and even without anything, we'd find the system tzdb.

So a bug related to this requires: no valid env vars (which is not the case in any of our build infra or user envs), an out-of-date system tzdb (<-- user's fault/problem IMO), stumbling over anything that actually changed in the tzdb compared to an older system tzdb (almost nothing changes unless a country changes its laws about timezones), _and_ using a timestamp just around the summer-/wintertime change. It's so extremely narrow a problem that I don't think the implementation effort for relative paths is anywhere near justified.

In any case, I'm trying to follow your and Isuru's preference. Help with the implementation is appreciated (so thanks for pointing out the bug) - I didn't expect it to be ready yet either, hence the debugging skips here. Right now I'm not even hitting the UAF due to the problems related to linking `libdl`, so that's the first thing to tackle.

h-vetinari:
I'm going to go with this for now:

> @isuruf: Let's remove the patch and merge and figure out tzdb stuff later.

I'll put up another PR with the WIP patch, and we can work from there, without indefinitely blocking the version upgrade.

h-vetinari:
Any objections to merging @isuruf?

recipe/meta.yaml
```diff
@@ -30,7 +30,7 @@ package:
   version: {{ version }}
 
 source:
-  - url: https://sourceware.org/pub/gcc/releases/gcc-{{ version }}/gcc-{{ version }}.tar.gz
+  - url: https://ftp.gnu.org/gnu/gcc/gcc-{{ version }}/gcc-{{ version }}.tar.gz
```
h-vetinari, line 33: FWIW, 12.4 wasn't available on the ftp server shortly after release, and the [announcement](https://gcc.gnu.org/pipermail/gcc/2024-June/244173.html) made it seem like that won't happen (also my [understanding](https://sourceware.org/sourceware-security-vision.html) of sourceware was that it's much closer to GCC than third-party). Just for context why I changed that.

isuruf: Ah, thanks. I didn't see the FSF affiliation.

h-vetinari:
@isuruf, AFAICT we have a bootstrapping problem here:
```
Could not solve for environment specs
The following package could not be installed
└─ gcc_impl_win-64 13.3.0.*  is not installable because it requires
   └─ libgcc-devel_win-64 13.3.0 h5200ebd_100, which requires
      └─ __win, which is missing on the system.
```

`libgcc-devel_win-64` is being built in that same job, but `gcc` (the output where things fail) wasn't part of the first mingw builds, so by the time it was enabled (https://github.com/conda-forge/ctng-compilers-feedstock/commit/e60f041730ef7292d9dfca6563e48df405485962) we already had some `libgcc` builds to work with.

I guess the subdir confusion is coming from a missing `build:` environment (even if it's empty, like it is for gcc_impl).

PR #134, opened 2024-05-14
Mingw - part 2

isuruf:
@conda-forge-admin, rerender.


recipe/meta.yaml
```diff
@@ -14,6 +14,13 @@
 {% set cross_target_platform = "linux-64" %}
 {% endif %}
 
+{% if cross_target_platform.startswith("win-") %}
+{% set runtime_suffix = "" %}
+{% else %}
+{% set runtime_suffix = "-ng" %}
```
h-vetinari, line 20: Longterm it would be nice to get rid of the runtime suffix everywhere; the last respective linux builds are:
* [libgcc](https://anaconda.org/conda-forge/libgcc/files): 5.2.0 (Aug. 2018)
* [libgfortran](https://anaconda.org/conda-forge/libgfortran/files?page=3): 3.0.0 (Aug. 2018)
* [libstdcxx](https://anaconda.org/conda-forge/libstdcxx/files): never published for linux

This should be fixable with some simple run-constraints (plus keeping `-ng` as a wrapper until we've switched over the strong run-exports for maybe a few years).

h-vetinari:
Re https://github.com/conda-forge/ctng-compilers-feedstock/pull/134/commits/d875cfd9edf003f2ffcd4b3af4514a6e8a04cb14, I don't think we need to work around the smithy bug. If the red X here bothers you, we should just cut a new smithy release (I didn't push for one because of other nearly finished PRs and not considering it very urgent).

isuruf:
Thanks. It'd be good to have a new release. I would've ignored a false positive, but this was failing conda-smithy, so I have to run conda-smithy locally every time.

h-vetinari:
> I would've ignored a false positive, but this was failing conda-smithy, so I have to run conda-smithy locally every time.

It should just fail the linter, not the rerender?

isuruf:
> It should just fail the linter, not the rerender?

Yes, I meant the linter.

h-vetinari:
GCC server looks like it's down. https://gcc.gnu.org/pipermail/gcc/ is affected too.

Issue #131, opened 2024-04-16
ppc64le: Issue with gcc-12 and gcc-13


RajalakshmiSR:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue
```
gcc -o t t.c -mcpu=native
cc1: error: unrecognized argument in option '-mcpu=native'
cc1: note: valid arguments to '-mcpu=' are: 401 403 405 405fp 440 440fp 464 464fp 476 476fp 505 601 602 603 603e 604 604e 620 630 740 7400 7450 750 801 821 823 8540 8548 860 970 G3 G4 G5 a2 cell e300c2 e300c3 e500mc e500mc64 e5500 e6500 ec603e power10 power3 power4 power5 power5+ power6 power6x power7 power8 power9 powerpc powerpc64 powerpc64le rs64 titan
```

### Installed packages

```shell
# packages in environment at /home/raji/anaconda3/envs/gccissue1:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
_sysroot_linux-ppc64le_curr_repodata_hack 4                   h43410cf_14    conda-forge
binutils_impl_linux-ppc64le 2.40                 hba8546e_0    conda-forge
bzip2                     1.0.8                ha17a0cc_5    conda-forge
ca-certificates           2024.2.2             h0f6029e_0    conda-forge
gcc                       13.2.0               hb8c6617_3    conda-forge
gcc_impl_linux-ppc64le    13.2.0               h7908001_5    conda-forge
kernel-headers_linux-ppc64le 3.10.0              h23d7e6c_14    conda-forge
ld_impl_linux-ppc64le     2.40                 hfa29eb9_0    conda-forge
libffi                    3.4.2                h4e0d66e_5    conda-forge
libgcc-devel_linux-ppc64le 13.2.0             hf1adbf4_105    conda-forge
libgcc-ng                 13.2.0               h31ea8bd_5    conda-forge
libgomp                   13.2.0               h31ea8bd_5    conda-forge
libnsl                    2.0.1                ha17a0cc_0    conda-forge
libsanitizer              13.2.0               h0849fd4_5    conda-forge
libsqlite                 3.45.3               hd4bbf49_0    conda-forge
libstdcxx-ng              13.2.0               h0849fd4_5    conda-forge
[...]

PR #142, opened 2024-06-28
tzdb integration

h-vetinari:
Retry bits that were de-scoped from #133


recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++    Dl_info info;
++
++    if (dladdr(addr, &info)) {
++      char* resolved = realpath(info.dli_fname, buffer);
++      if (resolved) {
++        int total_length = (int)strlen(resolved);
++        this_lib = (char*)malloc(total_length + 1);
++        memcpy(this_lib, resolved, total_length);
++#endif
++
```
jjerphan, line 76: For extra safety?
```suggestion
+
+        this_lib[static_cast<size_t>(total_length)] = '\0';
```

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++#   define TO_PREFIX "/.."
++#   define PATH_SEP '/'
++    char buffer[PATH_MAX];
++    int i;
++    Dl_info info;
++
++    if (dladdr(addr, &info)) {
++      char* resolved = realpath(info.dli_fname, buffer);
++      if (resolved) {
++        int total_length = (int)strlen(resolved);
```
jjerphan, line 72: Should we prefer `static_cast`?

This also applies in other places.

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++      // https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew#return-value
++      DWORD total_length = GetModuleFileNameW(hm, buffer, sizeof(buffer))
++      if (total_length) {
++        this_lib = (char*)malloc(total_length + 1);
++        memcpy(this_lib, buffer, total_length);
++#else
++    // we're in $PREFIX/lib/libstdcxx.so
++#   define TO_PREFIX "/.."
++#   define PATH_SEP '/'
++    char buffer[PATH_MAX];
```
jjerphan, line 65: Out of curiosity, where is `PATH_MAX` defined?

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++            break;
++          }
++        }
++        // <string> is included through <chrono>
++        std::string tz_dir(this_lib);
++        tz_dir += TO_PREFIX;
++        tz_dir += "/share/zoneinfo";
++        // std::string constructor deep-copies
++        free(this_lib);
++        return tz_dir.c_str();
```
jjerphan, line 90: `tz_dir` is local to this scope, so after this function returns the addressed previously returned by `tz_dir.c_str()` will be de-allocated.

I think the `char*` returned by `zoneinfo_dir`, must be `static` and allocated and set only once (when `zoneinfo_dir` is called).

recipe/build.sh
```diff
@@ -78,6 +78,8 @@ if [[ "$TARGET" == *linux* ]]; then
   GCC_CONFIGURE_OPTIONS+=(--enable-libsanitizer)
   GCC_CONFIGURE_OPTIONS+=(--enable-default-pie)
   GCC_CONFIGURE_OPTIONS+=(--enable-threads=posix)
+  # needs to link to libdl for finding tzdb (see patches)
```
jjerphan, line 81: Can you document which element of the patch requires `libdl`?

jjerphan (changes requested):
A first quick review.

recipe/build.sh
h-vetinari: the call to `dladdr`

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
h-vetinari: Yeah, this is what @mbargull pointed out as well in #131. I'll see if I can allocate statically in the function call.

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
h-vetinari: I took this over from https://github.com/gpakosz/whereami, but some googling says its from `<limits.h>`.

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
h-vetinari: It's impossible to have a conda environment where the `/lib` folder does not have at least depth 1 (and thus a guaranteed null byte that we'll insert below). Not sure what the extra null byte would achieve

h-vetinari:
OK, this now passes some basic tests, though I'm sure the implementation can be improved (e.g. less allocations as @mbargull pointed out in #133). It's still using the `-ldl` debug work-around (when we should really be adding `libdl` to the link interface of `libstdc++.so` (equivalent of CMake's `target_link_libraries`), but as I noted in #133:
> I _think_ it should be something like `AC_LIB_LINKFLAGS`, but the [makefile](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.1.0/libstdc%2B%2B-v3/src/Makefile.in)(s) in GCC are just indecipherable for me (though I really tried to find where the shared lib gets defined...). Any ideas how/where to do this?

CC @conda-forge/ctng-compilers

recipe/tests/tzdb.cpp
```diff
@@ -88,6 +88,8 @@ test_load_leapseconds()
 
 int main()
 {
+  const char* tz_dir = __gnu_cxx::zoneinfo_dir_override();
+  std::printf("using tzdb-path: %s\n", tz_dir);
```
h-vetinari, line 92: Hm. Perhaps it's not actually possible to access this weak symbol directly...?

h-vetinari: Nevermind, fixed it with an `extern` declaration.

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,103 @@
...
++
++        for (i = (int)total_length - 1; i >= 0; --i) {
++          if (this_lib[i] == PATH_SEP) {
++            // set to null byte so the string ends before the basename
++            this_lib[i] = '\0';
++            break;
++          }
++        }
++        // <string> is included through <chrono>
++        static std::string tz_dir(this_lib);
```
jjerphan, line 85: I think we better initialize it at the top of this function and return it directly if it was set (i.e. when this functions is subsequently called). What do you think?

h-vetinari: Yeah, sounds reasonable.

Right now I'm mostly wondering though why the linker cannot find `__gnu_cxx::zoneinfo_file`. It _does_ find `__gnu_cxx::zoneinfo_dir_override` (otherwise the CI for https://github.com/conda-forge/ctng-compilers-feedstock/pull/142/commits/7807b911f42533c59e6dcb69395b8f93a6115b36 wouldn't have passed), but I don't know why it cannot find its immediate neighbour. Possibly because of picking the wrong overload for the various `string_view` specializations between symbol creation and when it's called? 🤷

h-vetinari:
OK, so I made the variable static (also provides some caching if it has been loaded already), and managed to test this conclusively & successfully (see CI @ https://github.com/conda-forge/ctng-compilers-feedstock/pull/142/commits/108da784c0cfe886364b1e6306a75473ea9411dc)

I haven't managed to get rid of `-ldl` being for linking to `libstdc++.so`, I'm either going to need a tip there how to add this to the link interface through autotools, or alternatively, we could just add `-ldl` to the default C++ flags in the activation feedstock.

PTAL @isuruf @mbargull

recipe/patches/0006-add-ldl-to-libstdc___la_LDFLAGS.patch
```diff
@@ -0,0 +1,22 @@
...
+diff --git a/libstdc++-v3/src/Makefile.am b/libstdc++-v3/src/Makefile.am
+index 37ba1491d..55622ffe7 100644
+--- a/libstdc++-v3/src/Makefile.am
++++ b/libstdc++-v3/src/Makefile.am
+@@ -161,7 +161,7 @@ libstdc___darwin_rpath += -Wl,-rpath,@loader_path
+ endif
+ 
+ libstdc___la_LDFLAGS = \
+-	-version-info $(libtool_VERSION) ${version_arg} -lm $(libstdc___darwin_rpath)
++	-version-info $(libtool_VERSION) ${version_arg} -lm -ldl $(libstdc___darwin_rpath)
```
isuruf, line 19:
```suggestion
+	-version-info $(libtool_VERSION) ${version_arg} -lm -Bstatic -ldl -Bdynamic $(libstdc___darwin_rpath)
```

recipe/patches/0005-move-zoneinfo_file-function-into-__gnu_cxx-namespace.patch
```diff
@@ -0,0 +1,120 @@
+From 604f16b22f754741c8ac7a9b7dc13a97d3e0a15f Mon Sep 17 00:00:00 2001
+From: "H. Vetinari" <h.vetinari@gmx.com>
+Date: Mon, 1 Jul 2024 09:34:14 +1100
+Subject: [PATCH 5/6] move zoneinfo_file function into __gnu_cxx namespace
```
isuruf, line 4: Let's not move functions around like this. I don't think we should be changing the namespaces in libstdc++. If I read the code correctly, this is for testing?

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,111 @@
...
+-  const char* (*zoneinfo_dir_override)() = nullptr;
+-#else
+-  [[gnu::weak]] const char* zoneinfo_dir_override();
+-
+-#if defined(__APPLE__) || defined(__hpux__) \
+-  || (defined(__VXWORKS__) && !defined(__RTP__))
+-  // Need a weak definition for Mach-O et al.
+   [[gnu::weak]] const char* zoneinfo_dir_override()
+   {
++    // get path to library we're in, to determine our location relative to $PREFIX;
```
isuruf, line 44: Can we disable this for the static `libstdc++`? There should probably be a preprocessor directive that we can use for it.
When building executables with `-static-libstdc++` like micromamba, we shouldn't leave open a security hole like this.

recipe/patches/0005-move-zoneinfo_file-function-into-__gnu_cxx-namespace.patch
h-vetinari: There are more details in the commit message of https://github.com/conda-forge/ctng-compilers-feedstock/pull/142/commits/9ffd5185d5e5f89450f846f06b712b59de11e0de; without moving this function, we have no way to test our implementation of `zoneinfo_dir_override` (except implicitly through loading a timezone), because declaring `extern ... zoneinfo_dir_override` actually changes the effect that the function has, in the sense that the "root" we're using for the relative paths is not `libstdc++` anymore, but the TU where we're calling `zoneinfo_dir_override`.

recipe/tests/tzdb.cpp
```diff
@@ -88,6 +92,19 @@ test_load_leapseconds()
...
+  std::printf("using tzdb-path: %s\n", tz_dir.c_str());
+  const char* conda_prefix = getenv("CONDA_PREFIX");
+  if (conda_prefix) {
+    std::string expected(conda_prefix);
+    // this feedstock only compiles on linux, so this test is linux-only
+    // (since we need target_platform == cross_target_platform to execute it);
+    // on windows this would be "/Library/bin/../../share/zoneinfo"
+    expected += "/lib/../share/zoneinfo";
+    VERIFY( tz_dir == expected );
+  }
```
h-vetinari, line 106: In other words, we wouldn't be able to test (or even display for debugging) the path that our implementation ends up using like the above, unless we export another function that wraps `zoneinfo_dir_override`. Since `zoneinfo_file` does _exactly_ that, I think it's worth considering exporting that, for several reasons:
* it's a private namespace, so no-one is supposed to rely on it anyway
* being able to debug/display the value we end up using is useful
* it doesn't change the implementation, and is trivial except the namespace-change

h-vetinari (commented):
If we don't export a wrapper around `zoneinfo_dir_override`, it becomes essentially impossible to observe/test/print the values it produces.

recipe/patches/0004-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
h-vetinari: > There should probably be a preprocessor directive that we can use for it.

OK, I'll try to look around.

recipe/meta.yaml
```diff
@@ -325,6 +332,18 @@ outputs:
         - ${CXX} -Wall tests/hello_world.cpp -c -o hello_world.o --std=c++17
         - ${CXX} -Wall hello_world.o -o hello_world --std=c++17
 
+        # test tzdb integration
+      {% if gcc_maj_ver|int >= 14 %}
+        - ${CXX} -isystem ${PREFIX}/include --std=c++20 -Wall tests/tzdb.cpp -c -o tzdb.o
+        {% if target_platform == cross_target_platform %}
+        - ${CXX} tzdb.o -o tzdb && ./tzdb
+        - strace ./tzdb 2>&1 | grep ${CONDA_PREFIX}/lib/../share/zoneinfo/tzdata.zi
```
isuruf, line 340: What do you think about this test? This should be enough IMO

h-vetinari: Fine with me. 👍

h-vetinari:
> I haven't managed to get rid of `-ldl` being for linking to `libstdc++.so`, I'm either going to need a tip there how to add this to the link interface through autotools, or alternatively, we could just add `-ldl` to the default C++ flags in the activation feedstock.

@isuruf, this still fails when we compile something that needs `libstdc++`, because the `-ldl` is not automatically added. I don't know how to "bake this in"; unless you have an idea how to achieve this, I propose to add `-ldl` to the activation feedstock. WDYT?

isuruf:
> @isuruf, this still fails when we compile something that needs libstdc++, because the -ldl is not automatically added. I don't know how to "bake this in"; unless you have an idea how to achieve this, I propose to add -ldl to the activation feedstock. WDYT?

That's because `libstdc++.so` is not getting linked to `libdl` with my changes nor yours. Basically the patch you add doesn't work.

h-vetinari:
> Basically the patch you add doesn't work.

That's not a big surprise to me; it was an exploratory change, because I saw `-lm` in there, and hoped that it would similarly translate. I stared at the makefiles for a while, but I cannot even make out where the `.so` is defined, nor do I know for sure that `AC_LIB_LINKFLAGS` would be the right approach if I did find it.

recipe/patches/0006-add-ldl-to-libstdc___la_LDFLAGS.patch
```diff
@@ -0,0 +1,22 @@
...
+Subject: [PATCH 6/6] add `-ldl` to libstdc___la_LDFLAGS
+
+---
+ libstdc++-v3/src/Makefile.am | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/libstdc++-v3/src/Makefile.am b/libstdc++-v3/src/Makefile.am
+index 37ba1491d..55622ffe7 100644
+--- a/libstdc++-v3/src/Makefile.am
++++ b/libstdc++-v3/src/Makefile.am
```
mbargull, line 13: The build script doesn't run `autoreconf` or the like.
Try something like
```sh
sed -i.bak 's/^GLIBCXX_LIBS =.*/& -ldl/' libstdc++-v3/src/Makefile.in
```
or an equivalent patch, if preferred; a patch might be nicer, but IDK if it would induce more maintenance burden (depends on how frequently the context in the generated file changes).

h-vetinari:
> That's because `libstdc++.so` is not getting linked to `libdl` with my changes nor yours.

Though it _is_ getting linked successfully while building with that change; before this patch, I needed to explicitly add `-ldl` in `build.sh`; after the patch that was not necessary anymore. However, that linkage information doesn't persist until after the build, when we're trying to link the test code with libstdc++

recipe/patches/0006-add-ldl-to-libstdc___la_LDFLAGS.patch
```diff
@@ -0,0 +1,22 @@
+From b8b0bcd9d97f2dc59c566e0a836d3dc70ddd947b Mon Sep 17 00:00:00 2001
+From: "H. Vetinari" <h.vetinari@gmx.com>
+Date: Mon, 15 Jul 2024 19:16:05 +1100
+Subject: [PATCH 6/6] add `-ldl` to libstdc___la_LDFLAGS
+
+---
+ libstdc++-v3/src/Makefile.am | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/libstdc++-v3/src/Makefile.am b/libstdc++-v3/src/Makefile.am
```
isuruf, line 10: You need to edit `Makefile.in` file too. `.in` file is generated from `.am` file.

mbargull: (cross-posted something similar in https://github.com/conda-forge/ctng-compilers-feedstock/pull/142#discussion_r1680453673 )

h-vetinari: Ah OK; I thought the `.am` gets generated from the `.in` file?

h-vetinari:
If I do patch the Makefiles again, is it important for you to use the `-Bstatic -ldl -Bdynamic -Wl,--exclude-libs,libdl.a` stuff, or was that just exploratory as well?

recipe/patches/0006-add-ldl-to-libstdc___la_LDFLAGS.patch
mbargull: `*.am`/`*.ac` files are inputs for autotools ("ac": autoconf, "am": automake) to generate configure scripts and the template makefiles (`*.in`); configure scripts will then generate the final makefiles from the `*.in` files.

mbargull:
> If I do patch the Makefiles again, is it important for you to use the `-Bstatic -ldl -Bdynamic -Wl,--exclude-libs,libdl.a` stuff, or was that just exploratory as well?

I haven't yet thought about the static linking case; I'll leave this one for @isuruf since he is more of an expert than me on it ;).

isuruf:
Yes, we should link in `libdl.a` statically

recipe/tests/tzdb.cpp
h-vetinari: @isuruf @mbargull, would you prefer:
* not moving `zoneinfo_file` into `__gnu_cxx`, at the cost of losing the ability to test our implementation explicitly
* do the move (for the reasons above), and keep the tests

h-vetinari: Ah, nevermind, we covered this with the `strace` tests...

h-vetinari:
This should be good now. Thanks for your help @isuruf & @mbargull!

h-vetinari:
Gentle ping @conda-forge/ctng-compilers

h-vetinari:
All good now?

recipe/patches/0003-patch-zoneinfo_dir_override-to-point-to-our-tzdata.patch
```diff
@@ -0,0 +1,113 @@
...
+ 
+ #if defined __GTHREADS && ATOMIC_POINTER_LOCK_FREE == 2
+@@ -64,25 +68,73 @@
+ 
+ namespace __gnu_cxx
+ {
+-#ifdef _AIX
+-  // Cannot override weak symbols on AIX.
+-  const char* (*zoneinfo_dir_override)() = nullptr;
+-#else
```
isuruf, line 36: Let's keep this block or at least #error so that we don't trip if we build this.

isuruf:
Thanks

PR #148, opened 2024-08-06
Drop `-ng` suffix for runtime libraries, but keep compat-wrappers

h-vetinari:
This is a follow-up to a long-standing transition that has been in stasis. As noted in #134
> Longterm it would be nice to get rid of the runtime suffix everywhere; the last respective linux builds are:
> * [libgcc](https://anaconda.org/conda-forge/libgcc/files): 5.2.0 (Aug. 2018)
> * [libgfortran](https://anaconda.org/conda-forge/libgfortran/files?page=3): 3.0.0 (Aug. 2018)
> * [libstdcxx](https://anaconda.org/conda-forge/libstdcxx/files): never published for linux
> 
> This should be fixable with some simple run-constraints (plus keeping `-ng` as a wrapper until we've switched over the strong run-exports for maybe a few years).

The windows variants never introduced the suffix in the first place; drop it for the others too, but keep compatibility wrappers for a good while.


h-vetinari:
Draft due to rerendering errors: https://github.com/conda-forge/conda-smithy/issues/2012

h-vetinari:
This is ready now @conda-forge/ctng-compilers; I recommend to review commit by commit, the diff makes more sense that way.

h-vetinari:
For some reason, https://github.com/conda-forge/ctng-compilers-feedstock/pull/148/commits/ee7613f862746ff938c93e7dd46934364ec36b36 had fallen out of the last rebase - sorry.

isuruf (changes requested):
This would break all downstream recipes with `error_overlinking: true`. Probably needs a `run_exports` section in `libgcc-ng` to fix that.

isuruf:
We also need to run a mini-migrator that changes
```
ignore_run_exports:
  - libstdcxx-ng
```
to
```
ignore_run_exports_from:
  - {{ compiler('cxx') }}
```

h-vetinari:
> This would break all downstream recipes with `error_overlinking: true`. Probably needs a `run_exports` section in `libgcc-ng` to fix that.

My plan was to update the compiler activation as well of course. I don't think many recipes depend directly on `libgcc-ng` or `libgfortran`...? There's [5 feedstocks](https://github.com/search?q=org%3Aconda-forge+%2F%28%3Fm%29%28build%7Chost%29%3A%5Cs*%28%5C%23.*%29%3F%5Cn%28%5Cs*%5B%5C-%5C%23%5D.*%5Cn%29*%5Cs*-+lib%28gcc%7Cgfortran%7Cstdcxx%29-ng%2F+path%3Arecipe%2Fmeta.yaml&type=code) that have an explicit build or host dep on the `-ng` libraries (plus a handful that have a run-dep that should instead rely on the run-export). That said, I don't mind putting a run-export on `libgcc-ng` as an additional safety.

> We also need to run a mini-migrator [...]

Looking around, this only affects [about 20](https://github.com/search?q=org%3Aconda-forge+%2F(%3Fm)ignore_run_exports%3A\s*(\%23.*)%3F\n(\s*[\-\%23].*\n)*\s*-+libstdcxx-ng%2F+path%3Arecipe%2Fmeta.yaml&type=code) recipes. I think it would be less effort to fix them manually (I can do that) than write a migrator for all corner cases (and attach it to which migration...?).

h-vetinari:
OK, I've opened PRs for every feedstock that uses one of the `-ng` libraries [somewhere](https://github.com/search?q=org%3Aconda-forge+%2F-+lib%28gcc%7Cgfortran%7Cstdcxx%29-ng%2F+path%3Arecipe%2Fmeta.yaml&type=code&p=1) in meta.yaml, with the following exceptions:
* if `-ng` libraries are only used in `run`; those can/will be handled with a linter rule
* dead/archived feedstocks
  * https://github.com/conda-forge/go-gvisor-feedstock
  * https://github.com/conda-forge/gwframes-feedstock
  * https://github.com/conda-forge/heavydb-ext-feedstock
  * https://github.com/conda-forge/omniscidb-feedstock
  * https://github.com/conda-forge/tensorflow-estimator-feedstock

This leaves the following to-dos (not counting above-linked PRs that are green already):
  * [ ] [cuda-nvcc-impl](https://github.com/conda-forge/cuda-nvcc-impl-feedstock/blob/00a6db0bcb37601b012ba5e54f0a1697488bc0e0/recipe/meta.yaml#L99-L104) (should probably remove `-ng` suffix once we merge this PR)
  * [ ] [deno](https://github.com/conda-forge/deno-feedstock/blob/6640765e96be552662b0a60218ac4472c5662ac7/recipe/meta.yaml#L38-L40) (same)
  * [ ] https://github.com/conda-forge/gds-feedstock/pull/20 (broken on osx)
  * [ ] https://github.com/conda-forge/gst-python-feedstock/pull/58 (broken on windows)
  * [ ] https://github.com/conda-forge/tensorflow-hub-feedstock/pull/23 (broken)
  * [ ] https://github.com/conda-forge/uutils-coreutils-feedstock/pull/9 (broken on windows)

h-vetinari:
@isuruf, with the exception of ncurses & tzdata awaiting merge, all buildable feedstocks (i.e non-failing CI) are now free of `-ng` packages in build/host/ignore. For run-deps I'm going to add a lint-check, but this needs this PR first. Could you PTAL?

recipe/meta.yaml
```diff
@@ -637,6 +638,64 @@ outputs:
...
       license: GPL-3.0-only WITH GCC-exception-3.1
 
+  # compatibility outputs for previous naming of the runtime libraries
+  # with "-ng" suffix; for windows the suffix had never been introduced
+  {% if not cross_target_platform.startswith("win-") %}
+  - name: libgcc-ng
+    target: {{ cross_target_platform }}
+    build:
+      number: {{ build_num }}
+      skip: true    # [target_platform != cross_target_platform]
```
isuruf, line 648:
```suggestion
      skip: true    # [target_platform != cross_target_platform]
      run_exports:
        strong:
          - libgcc
```

isuruf: We need the same for the C++ and forrtran outputs.

recipe/meta.yaml
```diff
@@ -546,13 +549,8 @@ outputs:
...
         - msys2-conda-epoch <0.0a0    # [win]
-    build:
-      number: {{ build_num }}
-      detect_binary_files_with_prefix: False
-      skip: True   # [target_platform != cross_target_platform]
-      missing_dso_whitelist:
-        - "*"
-      rpaths_patcher: patchelf
+        # avoid installing incompatible version of old name for this output
+        - libgcc-ng =={{ version }}
```
isuruf, line 553:
```suggestion
        - libgcc-ng {{ version }}.* *_{{ build_num }}
```

isuruf: Otherwise we can have the old `libgcc-ng` with the same verison number interfere.

h-vetinari: Good point.

h-vetinari:
All good now?

isuruf:
Thanks

SylvainCorlay:
Hey @h-vetinari this seems to be causing clobbering between libgcc and libgcc-ng. Is this expected ?

SylvainCorlay:
If we have an environment with just libgcc-ng 14.1.0 and no libgcc, when a user modifies the environment it results in libgcc 14.1.0 being added, with the warning:

```
Warning  libmamba [libgcc-14.1.0-h77fa898_1] The following files were already present in the environment:
    - lib/libatomic.so
    - lib/libatomic.so.1
    - lib/libatomic.so.1.2.0
    - lib/libgcc_s.so
    - lib/libgcc_s.so.1
    - lib/libitm.so
    - lib/libitm.so.1
    - lib/libitm.so.1.0.0
    - lib/libquadmath.so
    - lib/libquadmath.so.0
    - lib/libquadmath.so.0.0.0
    - share/info/libgomp.info
    - share/info/libquadmath.info
    - share/licenses/gcc-libs/RUNTIME.LIBRARY.EXCEPTION
and then this results in all of those "colliding" files not being in the environment anymore!
```

h-vetinari:
> Hey @h-vetinari this seems to be causing clobbering between libgcc and libgcc-ng. Is this expected ?

Sorry about that. This shouldn't be possible based on the very tight run-constraints that we've added. Could you please open an issue with more details?

PR #198, opened 2025-11-25
Fix clobbering

isuruf:
Checklist

### DIFF
--- a/recipe/install-gcc.sh
+++ b/recipe/install-gcc.sh
@@ -224,7 +224,8 @@ if [[ "${HOST}" == "${TARGET}" ]]; then
 else
   source ${RECIPE_DIR}/install-libgcc.sh
   if [[ "${TARGET}" == *darwin* ]]; then
-    rm -rf ${PREFIX}/${TARGET}/lib/libgomp*
+    rm -rf ${PREFIX}/${TARGET}/lib/libgomp.*dylib
+    rm -rf ${PREFIX}/${TARGET}/lib/libgomp.a
   fi
   rm -f ${PREFIX}/share/info/*.info
   # TODO: create a TBD file libgomp.tbd that re-exports libomp.dylib


recipe/meta.yaml
```diff
@@ -341,6 +344,7 @@ outputs:
         # For cpp and crt{i,n}.o
         - {{ pin_subpackage("gcc_impl_" ~ cross_target_platform, exact=True) }}
         - binutils_impl_{{ cross_target_platform }} {{ binutils_version }}.* boostrap_*   # [cross_target_platform not in ("osx-arm64", "osx-64")]
```
h-vetinari: this is still using the wrong build string that got fixed in https://github.com/conda-forge/binutils-feedstock/pull/103 (and thus missing out on later changes like https://github.com/conda-forge/binutils-feedstock/pull/105)
```suggestion
        - binutils_impl_{{ cross_target_platform }} {{ binutils_version }}.* bootstrap_*   # [cross_target_platform not in ("osx-arm64", "osx-64")]
```

recipe/meta.yaml
h-vetinari: here as well (+ in the other places where binutils is used)

h-vetinari:
There's a bunch of failing jobs, for different reasons. The most benign one fails because it needs a build `osx-64/gcc_impl_linux-ppc64le-15.2.0-hd4c9bb7_12.conda` that [failed on main](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1400284&view=logs&j=8fe4044d-270e-5dde-6455-6d9927a029b1&t=6766e8fa-21e0-5edd-805e-93b46b6ebd95), though that build is passing now, so it'll work upon restarting after the merge once the first build has been uploaded.

Another (`osx-64/gcc_impl_win-64-15.2.0`) fails due to
```
error: $BUILD_PREFIX/bin/x86_64-apple-darwin13.4.0-install_name_tool: changing install names or rpaths can't be redone for: $PREFIX/bin/x86_64-w64-mingw32-gcov-tool (for architecture x86_64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
```

The majority is failing due to
```
+ test -f $PREFIX/lib/gcc/{{ target }}/15.2.0/libgomp.spec
```


Also worth noting is the following warning (for `osx-64/gcc_impl_osx-arm64-15.2.0-hdec9a96_13.conda`)
```
ld: warning: object file ($PREFIX/lib/gcc/arm64-apple-darwin20.0.0/15.2.0/libgcc.a(floatuntitd.o)) was built for newer macOS version (11.0) than being linked (10.13)
```
which indicates that something with the SDK/sysroot isn't being picked up correctly.

PR #205, opened 2025-12-25
Add support for RISC-V architecture

bitsk:
Checklist

This PR depends on merging [linux-sysroot-feedstock#98](https://github.com/conda-forge/linux-sysroot-feedstock/pull/98) first to resolve the RISC-V ABI directory issue. The GCC-13 build failed because the RISC-V `sysroot-with-crypt` is missing, as discussed https://github.com/conda-forge/ctng-compilers-feedstock/pull/127#discussion_r1573853024.


bitsk:

h-vetinari:
> The GCC-13 build failed because the RISC-V `sysroot-with-crypt` is missing, as discussed [here](https://github.com/conda-forge/ctng-compilers-feedstock/pull/127%23discussion_r1573853024).

That link is dead; however, there's no reason to build GCC 13 anymore[^1]. Just add a `skip: true  # [gcc_maj_ver == 13]`

[^1]: we're already on GCC 14 across conda-forge; that it's still built here is to keep things working for feedstocks that haven't transitioned to GCC 14 yet, but that's a very small number and for new architectures we're not constrained by any old builds

recipe/build.sh
```diff
@@ -140,6 +140,32 @@ if [[ ! ("${BUILD}" == "${HOST}" && "${HOST}" != "${TARGET}") && "${TARGET}" !=
   GCC_CONFIGURE_OPTIONS+=(--enable-lto)
 fi
 
+if [[ "$TARGET" == *riscv64* ]]; then
+  # GCC < 15 does not support the RVA23U64 profile.
+  # In this way, we can use GCC 14 to build rv64gc packages at first, 
+  # and when we switch to version 15 or later, 
+  # we can automatically switch to RVA23U64 and gradually replace the existing packages.
+  # The 'RVA23 Profiles' specification is available in
+  # https://drive.google.com/file/d/12QKRm92cLcEk8-5J9NI91m0fAQOxqNAq/view
```
isuruf: I don't think we'd want to build for RVA23U64 even in gcc 15. Those are new (non-existing) hardware and we don't want to keep using gcc 14 just to build for those new hardware. The way we do this in x86_64 is to build gcc for the baseline and use CFLAGS, CXXFLAGS etc when building specific package.

bitsk: Using CFLAGS and CXXFLAGS might not be suitable for all packages and could significantly increase the maintenance workload. It’s a bit of a dilemma at this point because RVA23 hardware is expected to hit the market in bulk by 2026.

We currently have a few options:

Build only rv64gc packages: This supports legacy hardware but lacks Bit-manipulation and Vector extensions (which are mandatory in aarch64), impacting performance for many packages.

Dual builds for both rv64gc and RVA23: This would double the required CI/build resources.

Gradual migration: Start with rv64gc and progressively transition to RVA23 (without a long-term commitment to legacy hardware).

Build RVA23 only: Focus exclusively on the new standard.

In short, we believe rv64gc-only hardware won't be around for the long term, and our distribution is already bottlenecked by its performance limitations. Given that new RISC-V chips will adopt the more mature RVA23 profile, I believe our RISC-V migration will align perfectly with the arrival of RVA23 hardware.

Which approach do you all prefer?

recipe/meta.yaml
```diff
@@ -222,7 +223,7 @@ outputs:
         - CC=$(${PREFIX}/bin/*-gcc -dumpmachine)-gcc
         - ${CC} -Wall tests/aligned_alloc.c -c -o c_aligned.o -v -Wl,-v -march=native  # [target_platform == cross_target_platform and (x86_64 or s390x)]
```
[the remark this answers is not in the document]
bitsk: RISC-V does not support the `-march=native` option.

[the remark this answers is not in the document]
isuruf: RISC-V doesn't support `-mcpu=native` either?

recipe/build.sh
isuruf: Build only rv64gc packages, but build RVA23 in specialized packages like OpenBLAS which can use runtime CPU feature detection.

h-vetinari: IOW, I don't want to end up in a situation where our longterm baseline is rv64gc; we need to be able to stay up-to-date and relevant with our packaging. If rv64gc will quickly become obsolete as @bitsk predicts, then we need to be able to move on from that.

Given that we currently have _**zero**_ users on riscv, I don't think it's worth our time to start with rv64gc only to remigrate to RVA23 after the initial riscv migration. But if people prefer that, I won't stand in the way.

recipe/meta.yaml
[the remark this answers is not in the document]
bitsk: Yes，RISC-V GCC doesn‘t support ` -mcpu=native`，See https://gcc.gnu.org/onlinedocs/gcc/RISC-V-Options.html

isuruf:
>  is there a need for a broader discussion on the RISC-V baseline policy?

There should be. Targeting a baseline just for the OpenKylin community who has new chips seems like a recipe for disaster. We need buy-in from a larger community of users and limiting ourselves to one company's chips is bad IMO.

recipe/build.sh
```diff
@@ -141,25 +141,12 @@ if [[ ! ("${BUILD}" == "${HOST}" && "${HOST}" != "${TARGET}") && "${TARGET}" !=
...
-      zifencei zihintntl zihintpause zihpm zimop \
-      zmmul za64rs zaamo zalrsc zawrs zfa \
-      zfhmin zca zcb zcd zcmop zba zbb \
-      zbs zkt zvbb zve32f zve32x zve64d \
-      zve64f zve64x zvfhmin zvkb zvkt zvl128b \
-      zvl32b zvl64b supm"
-    # Normalize whitespace then replace spaces with underscores: "i m a ..." -> "i_m_a_..."
-    RV_EXT="$(echo "${RV_EXT}" | tr -s '[:space:]' ' ' | sed -e 's/^ *//' -e 's/ *$//')"
-    GCC_CONFIGURE_OPTIONS+=(--with-arch="rv64${RV_EXT// /_}")
+  # According to discussions with core members, https://hackmd.io/@conda-community/SJ_oFitUZg
```
h-vetinari: Here's a more stable link to the minutes if you'd like:
https://github.com/conda/governance/blob/main/meetings/archive/20260204-conda-ecosystem.md

h-vetinari (approved):
Thanks, this LGTM. @isuruf, could you help unblock this please?

recipe/build.sh
```diff
@@ -140,6 +140,19 @@ if [[ ! ("${BUILD}" == "${HOST}" && "${HOST}" != "${TARGET}") && "${TARGET}" !=
   GCC_CONFIGURE_OPTIONS+=(--enable-lto)
 fi
 
+if [[ "$TARGET" == *riscv64* ]]; then
+  # According to discussions with core members, https://github.com/conda/governance/blob/main/meetings/archive/20260204-conda-ecosystem.md
+  # we have decided to continue using the rv64gc architecture in GCC 15, 
+  # and switch to the rva23u64 architecture in GCC 16 and later.
```
isuruf: Can you add the condition of rocky moving to `rva23u64` here?

h-vetinari:
Hey @bitsk, could you address the remaining review feedback here please? I suggest to leave out discussion and configuration of rva23 completely for now. We can continue with this topic in https://github.com/conda-forge/conda-forge.github.io/issues/1744 or a separate issue, rather than holding up this PR (this discussion ends up costing weeks that we could have been spending on building packages already).

I want us to get to actually kick off the migration and start building stuff, which will uncover a bunch of work that needs doing anyway, regardless of `rv64gc` vs `rva23`. Doing a rebuild later for a different baseline will then be only a fraction of the work compared to the initial roll-out.

bitsk:
@isuruf @h-vetinari  I've made the change, please take a look.

h-vetinari:
Not sure how this worked before, but we're not running into issues with [qemu](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/1fcb6141c7b8979c53d857117539280aff57591f/recipe/cross_compile_support.sh#L65) in the ci-setup

```
+++ /usr/bin/qemu-riscv64-static --version
/opt/conda/bin/cross_compile_support.sh: line 65: /usr/bin/qemu-riscv64-static: No such file or directory
```

Looks like we need to expand https://github.com/conda-forge/docker-images/blob/main/download-qemu-static.sh (or put work into https://github.com/conda-forge/docker-images/issues/320)

bitsk:
We need to merge this PR first.https://github.com/conda-forge/linux-sysroot-feedstock/pull/98

h-vetinari:
> We need to merge this PR first.[conda-forge/linux-sysroot-feedstock#98](https://github.com/conda-forge/linux-sysroot-feedstock/pull/98)

That PR has been merged now, but we're still missing qemu support for riscv

isuruf:
@conda-forge-admin rerender

bitsk:
@conda-forge-admin rerender

bitsk:
Hi, @isuruf, Let's hold off on building the native riscv64 toolchain for now and submit the cross-compilation toolchain first. We can compile the native toolchain after its dependencies have been successfully cross-compiled. I have already added the new package names to the admin-request https://github.com/conda-forge/admin-requests/pull/2019.

isuruf:
> Let's hold off on building the native riscv64 toolchain for now and submit the cross-compilation toolchain first.

Can you skip the native toolchains here and rerender?

bitsk:
Hi @isuruf , just wondering what the next step is to merge this PR?

isuruf:
Thanks. I'll restart the compilers once the cross-native compilers are built to see if the canadian cross compilers work.

luhenry:
:wave: Hello, following this PR, I don't see `gcc_linux-riscv64` in conda-forge [1], what is it that I'm missing?

[1]: https://anaconda.org/channels/conda-forge?pkg_q=gcc_linux-

PR #211, opened 2026-05-01
GCC v16.1.0, v15.3.0 & v14.4.0

h-vetinari:
Preconditions:
- [x] https://github.com/msys2/MINGW-packages/commit/2386f7b596fed897fa1f0dcdc56743251b47361c
  - see https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-gcc 
  - no dedicated patches for GCC 15.3 & 14.4 upstream yet; old v15.2 patches still apply & work for v15.3, but the same isn't true for v14 (though easily solved by https://github.com/conda-forge/ctng-compilers-feedstock/pull/211/commits/12e075f40faaffbe16a5f67f8eb05b8dce06eb95)
- [x] https://github.com/Homebrew/homebrew-core/pull/287916 based on upstream [branch](https://github.com/iains/gcc-16-branch); c.f. also [most recent list](https://github.com/Homebrew/homebrew-core/tree/main/Patches/gcc) of patches
- [x] https://github.com/iains/gcc-darwin-arm64/issues/153
  - worked around by https://github.com/conda-forge/ctng-compilers-feedstock/pull/211/commits/8a1a900b400f4584b45eb7fcb32a5cfe61985272
- [x] GCC v16 cross-compiling for mingw [doesn't find](https://github.com/conda-forge/ctng-compilers-feedstock/pull/211#issuecomment-5103796328) the C++ stdlib headers
  - fixed by https://github.com/conda-forge/ctng-compilers-feedstock/pull/211/commits/4f6b75ef66f1a4d0f393b1d5cf79d4d8fff69524


h-vetinari:
Something weird happening in conda-build with a symlink for a static library on v16; the build does
```
if test -f libatomic.a; then rm -f libatomic_asneeded.a; \
ln -s libatomic.a libatomic_asneeded.a; fi
```
and then later we fail _after_ building the output, somewhere in lief
```
Resource usage statistics from bundling gcc_impl_linux-64:
   Process count: 10
   CPU time: Sys=0:00:00.5, User=0:00:00.7
   Memory: 55.0M
   Disk usage: 7.4M
   Time elapsed: 0:00:18.1


number of files: 866
Making absolute symlink relative (libexec/gcc/x86_64-conda-linux-gnu/16.1.0/objdump -> $PREFIX/bin/x86_64-conda-linux-gnu-objdump :-> ../../../../bin/x86_64-conda-linux-gnu-objdump)

[... no mention of treating symlink for libatomic ...]

  File "/opt/conda/lib/python3.12/site-packages/conda_build/os_utils/liefldd.py", line 1191, in __call__
    with open(arg, "rb") as f:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '$PREFIX/lib/libatomic_asneeded.a'
```

jmakovicka:
> Something weird happening in conda-build with a symlink for a static library on v16; the build does
> 
> ```
> if test -f libatomic.a; then rm -f libatomic_asneeded.a; \
> ln -s libatomic.a libatomic_asneeded.a; fi
> ```
> 

`libatomic_asneeded` now needs to be moved together with `libatomic`
```
diff -ur recipe/install-gcc.sh parent/install-gcc.sh
--- recipe/install-gcc.sh       2026-05-04 04:21:54.568167346 -0400
+++ parent/install-gcc.sh       2026-01-19 11:16:39.000000000 -0500
@@ -227,7 +227,7 @@
 fi

 mkdir -p ${PREFIX}/lib/gcc/${TARGET}/${gcc_version}/
-for name in atomic gomp itm quadmath {a,hwa,l,t,ub}san; do
+for name in atomic_asneeded atomic gomp itm quadmath {a,hwa,l,t,ub}san; do
   if [[ -f "${PREFIX}/lib/lib${name}.a" ]]; then
    mv ${PREFIX}/lib/lib${name}.*a ${PREFIX}/lib/gcc/${TARGET}/${gcc_version}/
   fi
```

h-vetinari:
I'd be more tempted to delete `libatomic_asneeded.a`, rather than patching the (apparently non-functional) upstream build scripts. It didn't use to exist in gcc <16, so I'm not sure we really need it?

jmakovicka:
Apparently, the `asneeded` variants were added as a workaround for a libtool issue:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=123396

Which subsequently opened another can of worms on linker side:
https://gcc.gnu.org/pipermail/gcc-cvs/2026-January/448098.html

jmakovicka:
Note that I only had to change the line in the diff above (i.e. the last one for the static lib, not the dynamic one) for the build to pass.

h-vetinari:
Yeah, but I had seen in the build log and the bug you posted that there should also be a shared variant

```
if test -f libatomic.so; then (echo "/* GNU ld script"; \
 echo "   Add DT_NEEDED entry for -latomic only if needed.  */"; \
 echo "INPUT ( AS_NEEDED ( -latomic ) )" \
) > libatomic_asneeded.so; fi; \
```

I don't want to do this only half-way. Also, the failure is still the same; it looks like failure in conda-build, rather than during the build of GCC itself (or indeed our build scripts here).

recipe/install-gcc.sh
```diff
@@ -227,7 +227,7 @@ else
 fi
   
 mkdir -p ${PREFIX}/lib/gcc/${TARGET}/${gcc_version}/
-for name in atomic gomp itm quadmath {a,hwa,l,t,ub}san; do
+for name in atomic atomic_asneeded gomp itm quadmath {a,hwa,l,t,ub}san; do
```
jmakovicka, line 230: Ordering is important here. When you move `atomic` first, the `-f` test on the `atomic_asneeded ` symlink fails and the link will not be moved.

h-vetinari:
Regarding

https://github.com/conda-forge/ctng-compilers-feedstock/blob/b9e501aa1eea037b074788ec474f3f887fe4a03d/recipe/build.sh#L144-L151

I've checked 
* https://docs.rockylinux.org/release_notes/10_1/#support-for-risc-v-architecture
* https://rockylinux.org/news/rockylinux-support-for-riscv
* https://bugs.rockylinux.org/view_all_bug_page.php

and could not find any notice of the status of rva23 on current rocky (though v10.2 is scheduled to come out this month, I don't really expect changes of the architecture baseline in a patch release).

In the linked [meeting notes](https://github.com/conda/governance/blob/main/meetings/archive/20260204-conda-ecosystem.md) from February this year, @isuruf's suggested way forward was:
> Suggestion to start with rv64gc and GCC 15. Bump to rva23 and GCC 16 and new GLIBC when AlmaLinux does it too (e.g. next year). 

GCC 16 will not become our default compiler until mid-2027; I'd be OK with already defaulting to rva23 in GCC 16, that way riscv-feedstocks could also opt-in to a higher baseline where necessary (compared to gcc 15 which we're configuring for the older, slower rv64gc). Alternatively, we just stay on rv64gc for now, and we could still switch to rva23 for GCC 16 next year.

CC @conda-forge/help-riscv64

isuruf:
Let's keep rv64gc for gcc 16 for now. We can revisit that when rocky does too. Meeting notes should have been we move to rva23 with gcc 16 given that rocky moved to rva23.

h-vetinari:
> Meeting notes should have been we move to rva23 with gcc 16 given that rocky moved to rva23.

Even though the minutes are far from perfect, it's not feasible to edit them retrospectively ("actually, what's meant was ..."), that just leads to endless arguing. Rather, this should be rediscussed in a core call.

> We can revisit that when rocky does too. 

Actually, I think we may _have_ to rediscuss, because this is moving the goal posts in an impactful way (perhaps not vis-à-vis your interpretation, but at least compared to the fuzzy formulation in the minutes).

Rocky will almost certainly not move until RHEL 11, which has no release date yet, so it could be a few more years. I don't think the OpenKylin folks expected an indefinite hold, and we need to weigh their priorities (and thus willingness to contribute) as part of this process. It's taken half a year just to get compilers into place - I think we fell far short in terms of welcoming them to conda-forge, and so if anything we should be willing to compromise in their favour (rather than for hypothetical users on older riscv architectures, since we have no riscv64 packages at all yet).

fxcoudert:
Hi @h-vetinari thanks for reporting the cross-compiler GCC bug. I've filed it at https://github.com/iains/gcc-darwin-arm64/issues/153 and I'll follow up on that. As I stated in the linked Homebrew bug report, Homebrew will probably not ship a quick fix for that one, but I can help "vet" the fix for you, or help in any other way.

h-vetinari:
linux -> mingw for v15:
```
#include "..." search starts here:
#include <...> search starts here:
 $PREFIX/lib/gcc/x86_64-w64-mingw32/15.3.0/include
 $PREFIX/lib/gcc/x86_64-w64-mingw32/15.3.0/include-fixed
 $PREFIX/x86_64-w64-mingw32/sysroot/usr/include
End of search list.
Compiler executable checksum: 5307e505b7e24f74e3dba1228426319e
[... no error ...]
```

linux -> mingw for v16:
```
#include "..." search starts here:
#include <...> search starts here:
 $PREFIX/lib/gcc/x86_64-w64-mingw32/16.1.0/include
 $PREFIX/lib/gcc/x86_64-w64-mingw32/16.1.0/include-fixed
End of search list.
Compiler executable checksum: b47f113d7ebbaa5c7ce0beb1e46138e0
tests/hello_world.c:1:10: fatal error: stdio.h: No such file or directory
    1 | #include <stdio.h>
      |          ^~~~~~~~~
compilation terminated.
```

So it appears we're losing the sysroot (`$PREFIX/x86_64-w64-mingw32/sysroot/usr/include`) for v16 somewhere.

h-vetinari:
@conda-forge/ctng-compilers, this should be ready now, PTAL! :)

recipe/meta.yaml
```diff
@@ -26,41 +26,35 @@ source:
...
       - patches/mingw/{{ gcc_maj_ver }}/0003-Windows-Follow-Posix-dir-exists-semantics-more-close.patch
       - patches/mingw/{{ gcc_maj_ver }}/0005-Windows-Don-t-ignore-native-system-header-dir.patch
-      - patches/mingw/{{ gcc_maj_ver }}/0006-Windows-New-feature-to-allow-overriding.patch                      # [gcc_maj_ver == 13]
       - patches/mingw/{{ gcc_maj_ver }}/0007-Build-EXTRA_GNATTOOLS-for-Ada.patch
       - patches/mingw/{{ gcc_maj_ver }}/0008-Prettify-linking-no-undefined.patch
       - patches/mingw/{{ gcc_maj_ver }}/0011-Enable-shared-gnat-implib.patch
       - patches/mingw/{{ gcc_maj_ver }}/0012-Handle-spaces-in-path-for-default-manifest.patch
       - patches/mingw/{{ gcc_maj_ver }}/0014-gcc-9-branch-clone_function_name_1-Retain-any-stdcall-suffix.patch
       - patches/mingw/{{ gcc_maj_ver }}/0020-libgomp-Don-t-hard-code-MS-printf-attributes.patch
-      - patches/mingw/{{ gcc_maj_ver }}/0021-PR14940-Allow-a-PCH-to-be-mapped-to-a-different-addr.patch         # [gcc_maj_ver != 15]
```
isuruf, line 52: Any reason to remove this?

h-vetinari: The patch fails to apply; the error indicates that it has been applied upstream already, as I noted in https://github.com/conda-forge/ctng-compilers-feedstock/pull/211/commits/12e075f40faaffbe16a5f67f8eb05b8dce06eb95.

```
['-Np1', '-i', '/tmp/tmpnozzxmzq/0021-PR14940-Allow-a-PCH-to-be-mapped-to-a-different-addr.patch.native', '--binary']
checking file gcc/config/i386/host-mingw32.cc
Reversed (or previously applied) patch detected!  Skipping patch.
2 out of 2 hunks ignored
Warning: failed to download source.  If building, will try again after downloading recipe dependencies.
```

h-vetinari: As I outlined in the OP, there have been no mingw patches created specifically for 15.3 or 14.4 yet, so we're working off of those from `X.{Y-1}`; it's not too surprising IMO that a patch got upstreamed in the meantime.

isuruf (approved):
Looks good to me. Thanks

PR #219, opened 2026-07-29
Remove obsolete extra channels & a stray symlink

h-vetinari:
Follow-up to #211, there were some failure in the 2nd stage builds because of some old `gcc_impl_osx-64` [builds](https://anaconda.org/channels/conda-forge/packages/gcc_impl_osx-64/files?labels=gcc-experimental) under the `gcc-experimental` label which now clash with the current builds here in terms of channel priority; while we're at it, remove `sysroot-with-crypt`, which isn't needed anymore after we've dropped GCC 13, c.f.
https://github.com/conda-forge/ctng-compilers-feedstock/blob/f29301aeb71091e4b4a215a22c9e56d2a8378df1/recipe/setup_compiler.sh#L44
Finally, drop a symlink that causes failures in https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/200 (due to https://github.com/conda-forge/conda-smithy/issues/2638)

### DIFF
--- a/.ci_support/linux_64_cross_target_platformlinux-64cross_target_stdlibsysrootcross_target_stdlib_version2.17gcc_maj_ver14gcc_version14.4.0.yaml
+++ b/.ci_support/linux_64_cross_target_platformlinux-64cross_target_stdlibsysrootcross_target_stdlib_version2.17gcc_maj_ver14gcc_version14.4.0.yaml
@@ -9,7 +9,7 @@ c_stdlib:
 c_stdlib_version:
 - '2.17'
 channel_sources:
-- conda-forge/label/sysroot-with-crypt,conda-forge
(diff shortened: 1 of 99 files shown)


h-vetinari:
@conda-forge/ctng-compilers, this is ready. It's a necessary follow-up to #211; ppc-to-osx cross-compilers didn't build on main due to obsolete channels causing priority inversion, and a stray symlink causes failures in https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/200 and already a user bug report: #220 

For a more comprehensive guard against symlinks on windows we'd need to translate this recipe to v1, which we can do of course, but I don't want to block the version updates on this.

isuruf:
Thanks

PR #221, opened 2026-07-30
test upstream fix for mingw sysroot

h-vetinari:
In #211 we needed to "revert" https://github.com/msys2/MINGW-packages/commit/7d132b0972c81c65893eaa4a6e7b2e7436b2d6cf (see https://github.com/conda-forge/ctng-compilers-feedstock/commit/4f6b75ef66f1a4d0f393b1d5cf79d4d8fff69524); in order to avoid carrying changes of our own on top of the upstream patches, I commented about this in https://github.com/msys2/MINGW-packages/issues/27782; test https://github.com/msys2/MINGW-packages/pull/29280 as a proposed fix


h-vetinari:
Superseded by #222

